PR #12, opened 2019-09-15
Retry http PAR GETs

jeevanjoseph:
- switches to cURL for downloading the objects from the object storage.
- uses a single retry in case the initial request results in a non 200 response.
- cURL or wget does not retry 4xx errors even with the retry flags (those are for transient errors while an 404 is considered permanent)

### DIFF
--- a/deploy/basic/terraform/scripts/node.sh
+++ b/deploy/basic/terraform/scripts/node.sh
@@ -34,24 +34,54 @@ MUSHOP_APP_URI=$(curl -L http://169.254.169.254/opc/v1/instance/metadata | jq -j
 WALLET_URI=$(curl -L http://169.254.169.254/opc/v1/instance/metadata | jq -j ".wallet_par")
 
 # get artifacts from object storage
-wget -O /root/wallet.zip $${WALLET_URI}
-wget -O /root/catalogue.sql $${CATALOGUE_SQL_URI}
-wget -O /etc/httpd/conf/httpd.conf $${APACHE_CONF_URI}
-wget -O /root/entrypoint.sh $${ENTRYPOINT_URI}
-
+http_status=$(curl -w '%%{http_code}' -L -s -o /root/wallet.zip $${WALLET_URI})
+if [ "$http_status" != "200" ]; then
+    echo "Retrying $${WALLET_URI}"
+    sleep 10
+    curl -L -o /root/wallet.zip $${WALLET_URI}
+fi
 # Setup ATP wallet files
 unzip /root/wallet.zip -d /usr/lib/oracle/19.3/client64/lib/network/admin/
 
 # Init DB
+http_status=$(curl -w '%%{http_code}' -L -s -o /root/catalogue.sql $${CATALOGUE_SQL_URI})
+if [ "$http_status" != "200" ]; then
+    echo "Retrying $${CATALOGUE_SQL_URI}"
+    sleep 10
+    curl -L -o /root/catalogue.sql $${CATALOGUE_SQL_URI}
+fi
 sqlplus admin/$${ATP_PW}@$${ATP_DB_NAME}_tp @/root/catalogue.sql
 
+# Get http server config
+http_status=$(curl -w '%%{http_code}' -L -s -o /etc/httpd/conf/httpd.conf $${APACHE_CONF_URI})
+if [ "$http_status" != "200" ]; then
+    echo "Retrying $${APACHE_CONF_URI}"
+    sleep 10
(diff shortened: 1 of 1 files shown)


deploy/basic/terraform/scripts/node.sh
```diff
@@ -34,24 +34,54 @@ MUSHOP_APP_URI=$(curl -L http://169.254.169.254/opc/v1/instance/metadata | jq -j
...
-wget -O /root/catalogue.sql $${CATALOGUE_SQL_URI}
-wget -O /etc/httpd/conf/httpd.conf $${APACHE_CONF_URI}
-wget -O /root/entrypoint.sh $${ENTRYPOINT_URI}
-
+http_status=$(curl -w '%%{http_code}' -L -s -o /root/wallet.zip $${WALLET_URI})
+if [ "$http_status" != "200" ]; then
+    echo "Retrying $${WALLET_URI}"
+    sleep 10
+    curl -L -o /root/wallet.zip $${WALLET_URI}
+fi
```
mvandervliet: @jeevanjoseph these seem fine, but due to repetition, I might suggest a function like the following:

```bash
# load artifact from OS with retry
# @example
#   getArtifact $CATALOGUE_SQL_URI /root/catalogue.sql
getArtifact() {
  URI=$1
  OUT=$2
  http_status=$(curl -w "%{http_code}" -L -s -o $OUT $URI)
  echo "STATUS: $http_status"
  if [ "$http_status" != "200" ]; then
      echo "Retrying $URI"
      sleep 10
      getArtifact $1 $2
  fi
}
```

mvandervliet: ...(or adapted to only do single retry)

```bash
# load artifact from OS with retry
# @example
#   getArtifact $CATALOGUE_SQL_URI /root/catalogue.sql
getArtifact() {
  URI=$1
  OUT=$2
  http_status=$(curl -w "%{http_code}" -L -s -o $OUT $URI)
  echo "STATUS: $http_status"
  if [ "$http_status" != "200" ]; then
      echo "Retrying $URI"
      sleep 10
      curl -L -o ${OUT} ${URI}
  fi
}
```

junior: Let’s limit the number of retries, or would need a back off function, that would be unnecessary.

peterj: not sure what the timeout is in the whole thing, but it might make sense for us to exit 1 after a couple of unsuccessful retries

jeevanjoseph: working on it

jeevanjoseph: PR updated

deploy/basic/terraform/scripts/node.sh
```diff
@@ -34,24 +52,30 @@ MUSHOP_APP_URI=$(curl -L http://169.254.169.254/opc/v1/instance/metadata | jq -j
 WALLET_URI=$(curl -L http://169.254.169.254/opc/v1/instance/metadata | jq -j ".wallet_par")
```
junior: The `http://169.254.169.254/opc/v1/instance/metadata` should be a variable, as the usage repeats 5 times.

deploy/basic/terraform/scripts/node.sh
```diff
@@ -34,24 +52,30 @@ MUSHOP_APP_URI=$(curl -L http://169.254.169.254/opc/v1/instance/metadata | jq -j
...
 # Init DB
+get_object /root/catalogue.sql $${CATALOGUE_SQL_URI}
 sqlplus admin/$${ATP_PW}@$${ATP_DB_NAME}_tp @/root/catalogue.sql
 
-export OADB_USER=catalogue_user
-export OADB_PW=default_Password1
-export OADB_SERVICE=$${ATP_DB_NAME}_tp
+# Get http server config
+get_object /etc/httpd/conf/httpd.conf $${APACHE_CONF_URI}
+http_status=$(curl -w '%%{http_code}' -L -s -o /etc/httpd/conf/httpd.conf $${APACHE_CONF_URI})
```
mvandervliet: 🤔 This line should be removed

deploy/basic/terraform/scripts/node.sh
jeevanjoseph: Will raise as a separate PR, since that was not part of this change set

deploy/basic/terraform/scripts/node.sh
jeevanjoseph: Yep ! Good catch !

PR #13, opened 2019-09-17
deploy_assets_lifecycle_plus_fixes

junior:
Patch update.

Includes:
 - Object Storage lifecycle for deployment assets
 - README fixes
 - Version bumped to v1.0.1
 - catalogue cleanup
 - catalogue local zipkin

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,10 +15,10 @@ that creates all the required resources and configures the application on the cr
 
 The steps below guide you through deploying the application on your tenancy using the OCI Resource Manager.
 
-1. Download the latest [`mushop-v1.0.0-basic-stack.zip`](../../releases/download/v1.0.0/mushop-v1.0.0-basic-stack.zip) file.
-2. [Login](https://console.us-phoenix-1.oraclecloud.com/resourcemanager/stacks/create) to Oracle Cloud Infrastructure to import the stack
+1. Download the latest [`mushop-basic-stack-v1.0.1.zip`](../../releases/download/v1.0.1/mushop-basic-stack-v1.0.1.zip) file.
+2. [Login](https://console.us-ashburn-1.oraclecloud.com/resourcemanager/stacks/create) to Oracle Cloud Infrastructure to import the stack
     > `Home > Solutions & Platform > Resource Manager > Stacks > Create Stack`
-3. Upload the `mushop-v1.0.0-basic-stack.zip` file that was downloaded earlier, and provide a name and description for the stack
+3. Upload the `mushop-basic-stack-v1.0.1.zip` file that was downloaded earlier, and provide a name and description for the stack
 4. Configure the stack
    1. **Database Name** - You can choose to provide a database name (optional)
    2. **Node Count** - Select if you want to deploy one or two application instances.
--- a/deploy/basic/terraform/storage.tf
+++ b/deploy/basic/terraform/storage.tf
@@ -116,4 +116,16 @@ resource "oci_objectstorage_preauthrequest" "mushop_lite_preauth" {
   object = "${oci_objectstorage_object.mushop_basic.object}"
 }
 
+resource "oci_objectstorage_object_lifecycle_policy" "mushop_deploy_assets_lifecycle_policy" {
+  namespace = "${data.oci_objectstorage_namespace.user_namespace.namespace}"
+  bucket    = "${oci_objectstorage_bucket.mushop.name}"
 
+  rules {
+    action      = "DELETE"
+    is_enabled  = "true"
+    name        = "mushop-delete-deploy-assets-rule"
+    time_amount = "1"
+    time_unit   = "DAYS"
+
+  }
+}
--- a/src/catalogue/docker/docker-compose-zipkin.yml
+++ b/src/catalogue/docker/docker-compose-zipkin.yml
@@ -1,0 +1,52 @@
+#
+# Copyright (c) 2019, Oracle and/or its affiliates. All rights reserved.
+# Licensed under the Universal Permissive License v 1.0 as shown at http://oss.oracle.com/licenses/upl.
+# 
+
+version: '2'
+
+services:
+    catalogue:
+        build:
+            context: ..
+            dockerfile: ./Dockerfile
+        image: mushop/catalogue
+        hostname: catalogue
+        restart: always
+        cap_drop:
+            - all
+        cap_add:
+            - NET_BIND_SERVICE
+        read_only: true
+        volumes:
+            - Wallet_Creds:/usr/lib/oracle/19.3/client64/lib/network/admin/
+        environment:
+            - reschedule=on-node-failure
+            - OADB_USER=catalogue_user
+            - OADB_PW=default_Password1
+            - OADB_SERVICE=mcatalogue_tp
+            - ZIPKIN=http://zipkin:9411/api/v1/spans
+        ports:
+            - "8080:80"
+
+    zipkin:
+        image: openzipkin/zipkin 
+        hostname: zipkin
+        restart: always
+        cap_drop:
+            - all
+        cap_add:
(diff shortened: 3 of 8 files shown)


src/catalogue/docker/docker-compose-zipkin.yml
```diff
@@ -0,0 +1,52 @@
...
+        read_only: true
+        volumes:
+            - Wallet_Creds:/usr/lib/oracle/19.3/client64/lib/network/admin/
+        environment:
+            - reschedule=on-node-failure
+            - OADB_USER=catalogue_user
+            - OADB_PW=default_Password1
+            - OADB_SERVICE=mcatalogue_tp
+            - ZIPKIN=http://zipkin:9411/api/v1/spans
+        ports:
```
mvandervliet: Thinking we don’t actually want this file, do we?

src/catalogue/service.go
```diff
@@ -220,12 +220,3 @@ func cut(products []Product, pageNum, pageSize int) []Product {
 	}
 	return products[start:end]
 }
-
-func contains(s []string, e string) bool {
-	for _, a := range s {
-		if a == e {
-			return true
-		}
```
mvandervliet: Where this change coming from?

src/catalogue/docker/docker-compose-zipkin.yml
junior: It's the only use for [Zipkin](https://landscape.cncf.io/selected=zipkin) and [OpenTracing](https://landscape.cncf.io/selected=open-tracing) right now. This will keep evolving for the whole solution.

src/catalogue/service.go
junior: ?

junior: [wat](https://i.giphy.com/yjs4noW0nqYc8.mp4)

mvandervliet: Just that I don’t have the context for why it’s being removed? Deprecated?

src/catalogue/docker/docker-compose-zipkin.yml
mvandervliet: Well then maybe pull out pwds?

src/catalogue/service.go
junior: Old Function. Not in use. Part of the "catalogue cleanup".

src/catalogue/docker/docker-compose-zipkin.yml
junior: The pwd is a placeholder, also present on the docker-compose.yml that do the build locally. we can change to something more placeholdy, such as `change_it`

junior: Also is part of the .sql file.

src/catalogue/service.go
mvandervliet: 👌

src/catalogue/docker/docker-compose-zipkin.yml
junior: The name of the repo is xxx-cloudnative. How to use .env using the zip stack and ORM? The upcoming alternative deployment using K8s and OSB will generate all password automatically.

mvandervliet: Yes I am on the same page with these remarks, I just was not aware those features were being actively folded into this repository and the upcoming release.

As for `.env` in the basic/ORM deployment, some values are already set dynamically on init. `OADB_SERVICE` is dynamic today, but `OADB_PW` is hard-coded and should be converted to a generated value used when creating the schema. Is there a reason for hesitation with using `docker-compose` + `.env` now, when it's already possible?

junior: ORM uses the same sql file. If implement to set the password dynamically the TF script, then we can change all.

mvandervliet: I realize that... I'm just saying that it seems like a bad practice to _create_ tech debt when there's no reason to do so in this file.

junior: docker-compose.yml is already there. This just includes Zipkin for who want to do tracing/test with the service. The password placeholder is already present on the main use of deployment. By addressing this, need to change the TF script to not use the password placeholder for the limited-read-only DB user.

peterj: I would remove this zipkin file completely as it is not used anywhere at all. We are using source code control, there is history and we can always bring the zipkin compose file back if neededd.

mvandervliet: > docker-compose.yml is already there. This just includes Zipkin for who want to do tracing/test with the service. The password placeholder is already present on the main use of deployment. By addressing this, need to change the TF script to not use the password placeholder for the limited-read-only DB user.

Should be removed from the docker-compose.yml too. `mcatalogue_tp` is a completely different service anyway... changing these to use environment variables now doesn't mean any changes necessary to the TF script. You can leave the `CREATE USER...` statement as-is until TF can be updated to generate this. If these files are present, I'm only suggesting to move these values to an ignored `.env` file for the present time.

junior: Changing the entries to variables, matching the service usage. Not including new info for `.env`. It's up to the developer if wants to use EXPORT Variable or .env files.

jeevanjoseph: I think we can add this file back in when required.

mvandervliet: @peterj that all good by you?

README.md
```diff
@@ -15,10 +15,10 @@ that creates all the required resources and configures the application on the cr
 
 The steps below guide you through deploying the application on your tenancy using the OCI Resource Manager.
 
-1. Download the latest [`mushop-v1.0.0-basic-stack.zip`](../../releases/download/v1.0.0/mushop-v1.0.0-basic-stack.zip) file.
-2. [Login](https://console.us-phoenix-1.oraclecloud.com/resourcemanager/stacks/create) to Oracle Cloud Infrastructure to import the stack
+1. Download the latest [`mushop-basic-stack-v1.0.1.zip`](../../releases/download/v1.0.1/mushop-basic-stack-v1.0.1.zip) file.
+2. [Login](https://console.us-ashburn-1.oraclecloud.com/resourcemanager/stacks/create) to Oracle Cloud Infrastructure to import the stack
```
jeevanjoseph: good call !

deploy/basic/terraform/storage.tf
```diff
@@ -116,4 +116,16 @@ resource "oci_objectstorage_preauthrequest" "mushop_lite_preauth" {
   object = "${oci_objectstorage_object.mushop_basic.object}"
 }
 
+resource "oci_objectstorage_object_lifecycle_policy" "mushop_deploy_assets_lifecycle_policy" {
+  namespace = "${data.oci_objectstorage_namespace.user_namespace.namespace}"
+  bucket    = "${oci_objectstorage_bucket.mushop.name}"
+
+  rules {
```
jeevanjoseph: No object filters ? 
 - this will delete everything in the bucket, not an issue, just making sure that's what you intended.

junior: The bucket is only used for the deployment. The assets does not have any use after successfully deployed.

mvandervliet: No action required now, but any thoughts on how this might change once the catalogue images are also added to Object Storage?
- Separate bucket?
- Or use prefixes for `images` and `deploy` assets?

junior: Closing PR now. Separate buckets. That's the whole idea to have buckets, to separate the sand from the water to make the castle.

PR #36, opened 2019-10-31
Feature/terraformv12

jeevanjoseph:
- Stack is migrated to TF12
- AD auto selection (for free tier shapes)
- Infra layout changed to remove SVC GW and NAT GW

### DIFF
--- a/deploy/basic/terraform/main.tf
+++ b/deploy/basic/terraform/main.tf
@@ -10,7 +10,7 @@ data "template_file" "mushop" {
 
 resource "oci_core_instance" "app-instance" {
   count               = var.num_nodes
-  availability_domain = var.availability_domain
+  availability_domain = local.availability_domain[0]
   compartment_id      = var.compartment_ocid
   display_name        = "mushop-${random_id.mushop_id.dec}-${count.index}"
   shape               = local.instance_shape
@@ -65,7 +65,7 @@ locals  {
   
   instance_shape = "VM.Standard.E2.1.Micro"
 
-  availability_domain = 3
+  availability_domain = [for limit in data.oci_limits_limit_values.test_limit_values : limit.limit_values[0].availability_domain if limit.limit_values[0].value > 0 ]
 
   num_nodes = 2
 
--- a/deploy/basic/terraform/main.tf
+++ b/deploy/basic/terraform/main.tf
@@ -1,20 +1,23 @@
 # Copyright (c) 2019 Oracle and/or its affiliates. All rights reserved.
 # Licensed under the Universal Permissive License v 1.0 as shown at http://oss.oracle.com/licenses/upl.
 # 
+terraform {
+  required_version = ">= 0.12.0"
+}
 data "template_file" "mushop" {
   template = "${file("./scripts/node.sh")}"
 }
 
 resource "oci_core_instance" "app-instance" {
-  count               = "${var.num_nodes}"
-  availability_domain = "${var.availability_domain}"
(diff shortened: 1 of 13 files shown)


junior:
Please PR to Develop first

deploy/basic/terraform/schema.yaml
```diff
@@ -30,15 +29,6 @@ variables:
     title: "Compartment"
     description: "The compartment in which to create compute instance(s)"
 
-  availability_domain:
```
junior: Removed? The current command automatically get the Always Free resources AD?

deploy/basic/terraform/main.tf
```diff
@@ -62,7 +65,7 @@ locals  {
   
   instance_shape = "VM.Standard.E2.1.Micro"
 
-  availability_domain = 3
```
junior: What happen if the user has no limits on any AD? (e.g.: not on home region).

deploy/basic/terraform/main.tf
```diff
@@ -1,35 +1,38 @@
...
-    source_id   = "${local.images[var.region]}"
+    source_id   = local.images[var.region]
   }
 
-  metadata {
-    ssh_authorized_keys = "${var.ssh_public_key}"
-    user_data           = "${base64encode(data.template_file.mushop.rendered)}"
-    db_name             = "${oci_database_autonomous_database.mushop_autonomous_database.db_name}"
-    atp_pw              = "${random_string.autonomous_database_wallet_password.result}"
+  metadata = {
```
peterj: is this `=` a typo or?

deploy/basic/terraform/scripts/node.sh
```diff
@@ -52,8 +52,9 @@ MUSHOP_APP_URI=$(curl -L http://169.254.169.254/opc/v1/instance/metadata | jq -j
 WALLET_URI=$(curl -L http://169.254.169.254/opc/v1/instance/metadata | jq -j ".wallet_par")
 
 # get artifacts from object storage
-get_object /root/wallet.zip $${WALLET_URI}
+get_object /root/wallet.64 $${WALLET_URI}
 # Setup ATP wallet files
+base64 --decode /root/wallet.64 >  /root/wallet.zip
```
peterj: nit (sorry :)) you have an extra space after `>`

deploy/basic/terraform/schema.yaml
jeevanjoseph: yes

deploy/basic/terraform/main.tf
jeevanjoseph: deploy will fail

deploy/basic/terraform/main.tf
jeevanjoseph: No - this became a map key recently.

deploy/basic/terraform/scripts/node.sh
jeevanjoseph: Will fix

deploy/basic/terraform/main.tf
jeevanjoseph: can trials even switch regions ?

junior: yes. Another use case is that the user are on the home region, but is already using the Always-free resources

jeevanjoseph: When resources are already used, they will hit service limits. When they don’t have resources, ( like different region) they will get an error about the AD being invalid.

PR #45, opened 2019-11-21
Feature: documentation source

mvandervliet:
- Adds `src/docs` for producing `gh-pages` content
- Includes build and deploy instructions
- Preview build: https://mvandervliet.github.io/oci-cloudnative


mvandervliet:
> So, we still need to build the docs? That means that master will always be updated when there's a documentation change. We may use /docs instead of gh-pages then.

Yes... which, realistically, is the case no matter what. Ultimately it comes down to how the docs build is performed, and what that means in terms of commits. Consider the options:

1. **Using `/docs`**: Merge `master` -> build `src/docs` -> `mv src/docs/build docs/` -> commit to `master`
1. **Using `gh-pages`**: Merge `master` -> build `src/docs` -> push `gh-pages`

PR #52, opened 2019-12-02
Develop

junior:


peterj:
Could we move away from using develop and just simply use master? It might simplify things - we can still use other branches in case we need to do 'releases'.

PR #70, opened 2020-01-10
Feature/fix docker compose

jeevanjoseph:
updates to `docker-compose` with a deafult configuration that runs the app with mocked Orders, Catalogue and Users. Stream consumer is disabled (and can be enabled by the user.)


deploy/complete/docker-compose/docker-compose.yml
```diff
@@ -125,7 +129,7 @@ services:
     environment:
       - OADB_USER=orders_user
       - OADB_PW=Default_Password123#
-      - OADB_SERVICE=morders_tp
+      - OADB_SERVICE=mock
```
jeevanjoseph: This makes Orders run in mock mode, but the API server is not going to send any requests here since Users are mocked as well. When Users can run with a sqlite db, this will become relevant.

deploy/complete/docker-compose/docker-compose.yml
```diff
@@ -18,14 +18,16 @@ services:
...
-    image: iad.ocir.io/oracle/ateam/mushop-api:2.0.0
+    image: iad.ocir.io/oracle/ateam/mushop-api:2.0.1
     hostname: api
     restart: always
     cap_drop:
       - all
     cap_add:
       - NET_BIND_SERVICE
     read_only: true
+    ports:
```
jeevanjoseph: Exposing ports here to have the ability to directly hit the API without the edge router.

deploy/complete/docker-compose/docker-compose.yml
```diff
@@ -150,23 +154,23 @@ services:
...
-    read_only: true
-    environment:
-      - reschedule:on-node-failure
-      - OCI_TENANT_ID
-      - OCI_COMPARTMENT_ID
-      - OCI_REGION
-      - OCI_USER_ID
-      - OCI_FINGERPRINT
-      - OCI_API_KEY
+  # stream:
```
jeevanjoseph: Users can enable this. the default config is to be able to run locally, so this container would simply not be able to do that without the appropriate resources (stream) created.

deploy/complete/docker-compose/docker-compose.yml
```diff
@@ -105,12 +109,12 @@ services:
     environment:
       - OADB_USER=carts_user
       - OADB_PW=Default_Password123#
-      - OADB_SERVICE=mcarts_tp
+      - OADB_SERVICE=mock
```
jeevanjoseph: Carts in this configuration is live. the container will run and accept requests. Nothing is persisted though, since this the persistence is mocked (in memory).

deploy/complete/docker-compose/README.md
```diff
@@ -1,21 +1,50 @@
...
+
+- Users - not used, the API server provides a mock implementation that persists no data.
+- Orders - not used, the API server provides a mock implementation that persists no data.
+- Catalogue - not used, the API server provides a mock implementation that persists no data.
+- Carts - the `OADB_SERVICE=mock` configuration swaps out the Autonomous Database with an in memory one.
+- Stream - disabled and is not available in the offline mode.
+
+## Connecting to Oracle Cloud Infrastructure Services when running locally
+
+Where relevant, you can configure the container images to leverage cloud services as well. For instance, you can configure the locally runnin application to read and write data on to an Oracle ATP database.
```
mvandervliet: typo ~runnin~ running

deploy/complete/docker-compose/docker-compose.yml
mvandervliet: Can we just leave uncommented, since these vars are in the env pre-requisites anyway? One note about `shipping` and `stream` is that they should also include `STREAM_NAME`

deploy/complete/docker-compose/docker-compose.yml
mvandervliet: Created CNS-166 to support users in mock mode too. Can you set the same var in `user` service for future compatibility?

deploy/complete/docker-compose/README.md
```diff
@@ -1,21 +1,50 @@
...
+The complete **MuShop** application can be run using `docker-compose` locally.
+Like most 12-factor micro services, **MuShop** is built with testability 
+and portability in mind. The application configuration is [externalized](https://12factor.net/config) 
+from the container image so that the same application container image can be configured 
+to behave differently.  
 
-## Prerequisites
+Using `docker-compose` to run **MuShop** is useful when making changes to individual services or 
+running tests on the application locally. If you have a kubernetes cluster available, the helm 
+based deployment is recommended. Note that you can have a local kubernetes cluster based on [minikube](https://kubernetes.io/docs/tasks/tools/install-minikube/) or [kind](https://kind.sigs.k8s.io/)
```
mvandervliet: I'm inclined to suggest we omit this introduction altogether.

mvandervliet:
> The stream is disabled (commented) because I would like the users to un-comment it if they have setup a stream. This entire container may be redundant if we remove the stream component from here and move to the Orders --> message/nats.io --> Fulfillment (new service)

wouldn't the `shipping` service require the same then? When `orders` is in mock mode, does it still try calling `shipping`/`payment`?

deploy/complete/docker-compose/README.md
jeevanjoseph: I thought that we needed some context to reinforce that the container images are the same whether the user chooses the helm based installation or this. However the main reason someone would want to use this is to test quickly.

I am okay to remove this if you still think its unnecessary. The whole intro and the Prereq stuff or just the new content under prereq?

deploy/complete/docker-compose/docker-compose.yml
```diff
@@ -85,12 +89,12 @@ services:
     environment:
       - reschedule=on-node-failure
       - OADB_USER=catalogue_user
-      - OADB_PW=default_Password1
+      - OADB_PW
       - OADB_SERVICE=mcatalogue_tp
```
mvandervliet: should be "mock" as well, for consistency

deploy/complete/docker-compose/docker-compose.yml
```diff
@@ -188,6 +192,6 @@ services:
       - NET_BIND_SERVICE
     read_only: true
     environment:
-      - OADB_USER
+      - OADB_USER=userdb_admin
```
mvandervliet: no need for admin here

jeevanjoseph: The existing convention is _user. In this case that would mean users_user. that sounded confusing, which is why I chose this. Will fix.

PR #71, opened 2020-01-13
Feature: storefront alternate version

mvandervliet:
Supports a build-arg override to `VERSION`, where this information is passed into the UI static build. Version information is surfaced in the app footer. 

If `VERSION` matches the `/beta/i` regex, then a banner will be shown on the homepage indicating "BETA" preview:

image

image


src/storefront/README.md
```diff
@@ -73,10 +73,21 @@ make down
...
+Standard build
+
 ```shell
 docker build -t mushop/storefront .
 ```
 
+The storefront also supports **BETA** build option where an override to the 
+service semver can be provided as a build argument. When the supplied version 
+matches `/beta/i`, a banner is shown in the storefront UI. This is used for the
+purpose of demonstrating service mesh
```
peterj: any reason why is this passed through the build? it could be a simple environment variable set on the yaml.

mvandervliet: The build is just static HTML and an `nginx` static server (it's not doing any environment variable interpolation in the `nginx.conf` on startup either). So, it's nice to be able to do verifications with curl like:

```shell
curl -s -X GET localhost:80 | grep version
 <div class="version">v2.0.2-beta</div>
```

An alternative would be to use `env` interpolation in the config, then always respond with an `x-app-version` header or something

PR #72, opened 2020-01-15
Add newsletter-subscription function

peterj:
This PR adds a simple function that gets invoked (through an API gateway) when users subscribe to the newsletter. Note that the subscription is not managed anywhere and function is merely sending an email message to the provided address.

The function is invoked through an API gateway, so we don't need to deal with signing and it allows us to add auth/cors/rate limiting later if needed. 

The full URL to the function is provided to the `api` service through the`NEWSLETTER_SUBSCRIBE_URL` environment variable. This is not a required variable and if it not provided, the `api` service will just respond with a 200. Note that originally had the mock mode for the function, however, since we don't have a default value for the URL, like we do for other services (e.g. `http://carts`), it didn't make sense to spend time to re-work the mock mode detection and to default to mock mode if env values were not provided (and I don't think that would be great solution either).


deploy/complete/helm-chart/README.md
```diff
@@ -176,6 +176,20 @@ These services are configured using kubernetes secrets.
...
+If you want to configure and use the functions/API Gateway functionality, make sure you create and deploy the function and the API Gateway by following the instructions in the `src/functions/newsletter-subscription` folder.
+
+To configure the `api` chart to use the newsletter subscribe function, add the following snippet to the values file you are planning on deploying:
+```
+api:
+  env:
+    newsletterSubscribeUrl: https://[API_GATEWAY_URL]
+```
+
+Replace the `API_GATEWAY_URL` with an actual URL you got when you deployed your API gateway instance.
```
junior: We should point to how to deploy API gateway.

src/functions/newsletter-subscription/package.json
```diff
@@ -0,0 +1,13 @@
+{
+    "name": "newsletter-subscription",
+    "version": "0.1.0",
+    "description": "MuShop Newsletter Subscription",
+    "main": "func.js",
+    "author": "",
+    "license": "Apache-2.0",
```
junior: We may need to change to UPL

deploy/complete/helm-chart/README.md
peterj: I mention that in the functions README.md.

src/functions/newsletter-subscription/package.json
peterj: This was the default - who do we need to check with to change it?

peterj: Looks like the api/storefront `package.json` doesn't have the license defined, so we probably fallback to whatever the repo has. Will remove it.

src/functions/newsletter-subscription/func.js
```diff
@@ -0,0 +1,77 @@
...
+fdk.handle(function (input) {
+    if (!input.email) {
+        return { 'error': 'email not provided' };
+    }
+
+    const user = process.env.SMTP_USER;
+    const pass = process.env.SMTP_PASSWORD;
+    const host = process.env.SMTP_HOST;
+    const port = process.env.SMTP_PORT;
+    const approvedSender = process.env.APPROVED_SENDER_EMAIL;
```
mvandervliet: nit: could be a one-liner:

```js
const { SMTP_USER,  SMTP_PASSWORD, SMTP_HOST, SMTP_PORT, APPROVED_SENDER_EMAIL } = process.env;
```

src/storefront/src/scripts/shop/components/newsletter.js
```diff
@@ -0,0 +1,37 @@
+/**
+ * Copyright © 2019, Oracle and/or its affiliates. All rights reserved.
+ * The Universal Permissive License (UPL), Version 1.0
+ */
+import { Mu, MuMx, attrToSelector } from '../../mu';
```
mvandervliet: So nice.. this `Mu` thing is awesome 😉

peterj: yeah :) debateable :D

src/functions/newsletter-subscription/README.md
```diff
@@ -0,0 +1,121 @@
...
+
+Next, you need to generate the SMTP credentials that will allow you to log in to the SMTP server and send the email. Follow the [Generate SMTP Credentails for a User](https://docs.cloud.oracle.com/iaas/Content/Email/Tasks/generatesmtpcredentials.htm) to get the SMTP host, port, username and password.
+
+The SMTP credentails (host, port, username and password) and the approved sender email address (e.g. `mushop@example.com`) will be provided to the function as configuration values later, so make sure you save these values somewhere.
+
+## Create the application
+
+Each function needs to live inside of an application. You can create a new application either through the console, API or the Fn CLI. An application has a name (e.g. `mushop-app`) and the VCN and a subnet in which to run the functions. The one guideline here is to pick the subnets that are in the same region as the Docker registry you specified in your context YAML earlier.
```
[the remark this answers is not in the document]
peterj: Add the link.

src/functions/newsletter-subscription/func.js
```diff
@@ -41,12 +41,7 @@ fdk.handle(function (input) {
         return { 'error': 'email not provided' };
     }
 
-    const user = process.env.SMTP_USER;
-    const pass = process.env.SMTP_PASSWORD;
-    const host = process.env.SMTP_HOST;
-    const port = process.env.SMTP_PORT;
-    const approvedSender = process.env.APPROVED_SENDER_EMAIL;
-
+    const { user, pass, host, port, approvedSender } = process.env;
```
mvandervliet: 🤔 Are you also changing the function configuration with these keys? As far as consistency goes, I'd tend to favor the `CAPS` env vars...

peterj:
ugh, missed that - I’ll fix it

PR #73, opened 2020-01-15
Feature/carts updates

jeevanjoseph:
A set of minor improvements for carts

- Health checks that include db connectivity.
- App lifecycle is decoupled from DB initialization. App will stay up even if the DB initialization fails (app's health check will report unhealthy state)
- Access logging for the webserver
- Some application level logging


src/carts/pom.xml
```diff
@@ -15,7 +15,7 @@
     <properties>
         <maven.compiler.source>1.8</maven.compiler.source>
         <maven.compiler.target>1.8</maven.compiler.target>
-        <helidon.version>1.3.1</helidon.version>
+        <helidon.version>1.4.1</helidon.version>
         <jsonb.version>1.0.2</jsonb.version>
         <yasson.version>1.0.5</yasson.version>
```
junior: Why not bump version of yasson and jsonb?

src/carts/src/main/java/mushop/carts/Main.java
```diff
@@ -3,8 +3,11 @@
 import io.helidon.config.Config;
 import io.helidon.health.HealthSupport;
 import io.helidon.health.checks.HealthChecks;
+import org.eclipse.microprofile.health.HealthCheck;
```
junior: mp?

src/carts/src/main/java/mushop/carts/CartRepositoryDatabaseImpl.java
```diff
@@ -141,5 +146,18 @@ public void save(Cart cart) {
...
         }
     }
+
+    @Override
+    public boolean healthCheck(){
+        try (Connection con = pool.getConnection()) {
+            OracleDatabase db = SODA.getDatabase(con);
+            OracleCollection col = db.openCollection(collectionName);
+            String name = col.admin().getName();
+            return name == null ? false : true;
```
junior: Health check just return true or false? no payload? This case does not return health check for the application?

peterj: you could probably just do `return name != null;`

src/carts/src/main/java/mushop/carts/CartService.java
```diff
@@ -14,159 +17,192 @@
...
-                    carts.save(newCart);
-                    response.status(201).send(); // created
-                } else {
-                    cart.merge(newCart);
-                    carts.save(cart);
-                    response.status(200).send(); // ok
+                try {
+                    Cart cart = carts.get().getById(cartId);
+                    if (cart == null) {
+
```
peterj: nit: extra empty line :)

src/carts/src/main/java/mushop/carts/CartService.java
```diff
@@ -14,159 +17,192 @@
...
+                    Cart cart = carts.get().getById(cartId);
+                    if (cart == null) {
+                        response.status(404).send();
+                        return;
+                    }
+                    for (Item item : cart.getItems()) {
+                        if (item.getItemId().equals(qItem.getItemId())) {
+                            item.setQuantity(qItem.getQuantity());
+                            carts.get().save(cart);
+                            response.status(200).send();
```
peterj: you can probably return here after the `.send()` so it's not iterating through the cart items anymore.

src/carts/src/main/java/mushop/carts/Main.java
jeevanjoseph: No. Not using MP, but these custom health checks are in the MP package, but made available through io.helidon.health:helidon-health jars, which seemingly don't need MP. 

https://helidon.io/docs/latest/#/health/01_health

src/carts/src/main/java/mushop/carts/CartService.java
jeevanjoseph: Yeah - I noticed this too - I think the 
```
carts.get().save(cart);
response.status(200).send();
```
needs to be outside the loop.

this is the original logic here, and I will fix this in a separate PR

src/carts/src/main/java/mushop/carts/CartRepositoryDatabaseImpl.java
jeevanjoseph: fixed

src/carts/src/main/java/mushop/carts/CartService.java
jeevanjoseph: fixed

src/carts/src/main/java/mushop/carts/CartRepositoryDatabaseImpl.java
jeevanjoseph: Helidon does support adding a payload, but the method that returns the boolean is simply used to construct a standard payload (with just status messges.) The Builder for the healthcheck will use the method to determine the up/down status and then a separate paylaod method to create payload. There wasn't really any thing here that I could use for a a better payload. It does however use JVM metrics for the health as well in addition to the "custom" DB healthcheck. the response from /health will look like this: 
The states and logical `AND`ed to arive at the overall health, which also determines the HTTP response  code (200=OK, 503=unhealthy, 500 = healthcheck itself had some error, so app status is unknown) 
```
{
    "outcome": "UP",
    "status": "UP",
    "checks": [
        {
            "name": "dbHealth",
            "state": "UP",
            "status": "UP"
        },
        {
            "name": "deadlock",
            "state": "UP",
            "status": "UP"
        },
        {
            "name": "diskSpace",
            "state": "UP",
            "status": "UP",
            "data": {
                "free": "19.35 GB",
                "freeBytes": 20781621248,
                "percentFree": "33.13%",
                "total": "58.41 GB",
                "totalBytes": 62722478080
            }
        },
        {
            "name": "heapMemory",
            "state": "UP",
            "status": "UP",
            "data": {
                "free": "23.14 MB",
                "freeBytes": 24260408,
                "max": "128.00 MB",
                "maxBytes": 134217728,
                "percentFree": "68.08%",
                "total": "64.00 MB",
                "totalBytes": 67108864
            }
        }
    ]
}
```

src/carts/pom.xml
jeevanjoseph: done

src/carts/src/main/java/mushop/carts/CartService.java
```diff
@@ -14,159 +17,199 @@
...
     public void deleteCart(ServerRequest request, ServerResponse response) {
         try {
-            if (carts.deleteCart(request.path().param("cartId"))) {
+            if (carts.get().deleteCart(request.path().param("cartId"))) {
                 response.status(200).send();
             } else {
                 response.status(404).send();
             }
         } catch (Exception e) {
+            log.log(Level.SEVERE,"deleteCart failed." ,e);
```
peterj: nit: space should be after `,` not before

src/carts/src/main/java/mushop/carts/CartService.java
```diff
@@ -14,159 +17,199 @@
...
-                    carts.save(cart);
-                    response.status(200).send(); // ok
+                try {
+                    Cart cart = carts.get().getById(cartId);
+                    if (cart == null) {
+                        newCart.setId(cartId);
+                        carts.get().save(newCart);
+                        response.status(201).send(); // created
+                    } else {
+                        log.info("Existing Cart"+cart);
```
peterj: are these debug statements?

src/carts/src/main/java/mushop/carts/CartService.java
```diff
@@ -14,159 +17,199 @@
...
+                    if (cart == null) {
+                        response.status(404).send();
+                        return;
+                    }
+                    for (Item item : cart.getItems()) {
+                        if (item.getItemId().equals(qItem.getItemId())) {
+                            item.setQuantity(qItem.getQuantity());
+                            carts.get().save(cart);
+                            response.status(200).send();
+                            break;
```
peterj: probably needs return here, so the `response.status(404).send();` is not executed

src/carts/src/main/java/mushop/carts/Item.java
```diff
@@ -50,4 +50,36 @@ public void setUnitPrice(BigDecimal unitPrice) {
...
+    public int hashCode() {
+        final int prime = 31;
+        int result = 1;
+        result = prime * result + ((itemId == null) ? 0 : itemId.hashCode());
+        return result;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj)
```
peterj: another nit: use `{ }` :)

mvandervliet: 👍

src/carts/src/main/java/mushop/carts/Cart.java
```diff
@@ -50,9 +51,24 @@ public boolean removeItem(String itemId) {
 
     public void merge(Cart cart) {
         this.customerId = cart.getCustomerId();
-        for (Item item : cart.items) {
-            items.add(item);
-        }
+        List<Item> existingItems = new ArrayList<Item>(this.items);
+        cart.items.forEach((item) -> {
```
jjspiegel: Thanks for fixing the bugs and sorry to snoop on your change. 

I assume the fix here is to avoid duplicate items.  i.e (1) if the item already exists, then update its quantity and (2) if the item id does not exist, add it.  

Currently it is making a copy of the existing items before iterating the incoming items.  For each item, you are calling contains() which for ArrayList is going to iterate the internal array to test if its there.  Then, you will iterate the existingItems again when you call forEach.   Also, so that the contains has has the correct semantics, you have added hashCode and equals to Item.  (btw, ArrayList doesn't use hashCode :
https://docs.oracle.com/javase/8/docs/api/java/util/List.html#contains-java.lang.Object- )

It just occurred to me that you could avoid hashCode, equals, the copy, and these extra iterations if you do something like this:

```
public void merge(Cart cart) {
   for (Item item : cart.items) {
      mergeItem(item);
   }
}

private void mergeItem(Item item) {
   for (Item existing : items) {
      if (existing.getId() == item.getId()) {
         existing.setQuantity(.....);
         return;
      }
   }
   existing.add(item);
}
```

Also, should it be existing.getQuantity() + item.getQuanity()?

jeevanjoseph: Yes thats concise and much more efficient ! Thanks !
I'm updating the PR to use this

jeevanjoseph: BTW - I dont think you are in this github org? I could not assign you as a reviewer.

jeevanjoseph: @jjspiegel - Actually it will result in concurrent modification. The same thread that is iterating over the List in `merge` would be trying to add to the List in `mergeItem`, and that would throw it off with a runtime exception.

May be a better way is to not use `.forEach` and do a very traditional indexed iteration, since these are arraylists and index access is more efficient.

src/carts/src/main/java/mushop/carts/CartService.java
jeevanjoseph: Yes - they were. fixed now.

src/carts/src/main/java/mushop/carts/Cart.java
```diff
@@ -49,10 +50,24 @@ public boolean removeItem(String itemId) {
     }
 
     public void merge(Cart cart) {
-        this.customerId = cart.getCustomerId();
```
jjspiegel: Why remove this.customerId = cart.getCustomerId()?
I thought this handles case where user starts a cart before logging in and then logs in and starts adding more items.

PR #74, opened 2020-01-16
Feature users mockdb

mvandervliet:
Adds support for in memory database when `process.env.OADB_SERVICE === 'mock'`. This will allow the deprecation of `api/mock` middleware duplicating microservice functionality as well as mock mode co-dependencies (like `users <- orders`)


mvandervliet:
> The users mock db will pre-load mock data or will just be another in-memory db?

Not seeded, just the service... I'd suggest a separate seeding script if that's desired

PR #80, opened 2020-01-27
Add API Gateway, function and Email Delivery sections to the docs

peterj:


src/docs/source/includes/_deployment.md
```diff
@@ -246,6 +246,103 @@ values as described [above](#provisioning)
     kubectl get servicebindings -A
     ```
 
+## API Gateway, OCI Functions and Email Delivery
```
mvandervliet: Does this length work ok in the navigation?

src/docs/source/includes/_deployment.md
```diff
@@ -246,6 +246,103 @@ values as described [above](#provisioning)
...
+
+The SMTP credentails (host, port, username and password) and the approved sender email address (e.g. `mushop@example.com`) will be provided to the function as configuration values later, so make sure you save these values somewhere.
+
+### Configure function applicaiton
+
+Each function needs to live inside of an application. You can create a new application either through the console, API or the Fn CLI. An application has a name (e.g. `mushop-app`) and the VCN and a subnet in which to run the functions. The one guideline here is to pick the subnets that are in the same region as the Docker registry you specified in your context YAML earlier - check these [docs](https://docs.cloud.oracle.com/iaas/Content/Functions/Tasks/functionscreatingapps.htm) for more information.
+
+To create an application using Fn CLI, run:
+
+```
```
mvandervliet: Add `shell` or `text` language to the block

src/docs/source/includes/_deployment.md
```diff
@@ -246,6 +246,103 @@ values as described [above](#provisioning)
...
+
+Next, you need to generate the SMTP credentials that will allow you to log in to the SMTP server and send the email. Follow the [Generate SMTP Credentails for a User](https://docs.cloud.oracle.com/iaas/Content/Email/Tasks/generatesmtpcredentials.htm) to get the SMTP host, port, username and password.
+
+The SMTP credentails (host, port, username and password) and the approved sender email address (e.g. `mushop@example.com`) will be provided to the function as configuration values later, so make sure you save these values somewhere.
+
+### Configure function applicaiton
+
+Each function needs to live inside of an application. You can create a new application either through the console, API or the Fn CLI. An application has a name (e.g. `mushop-app`) and the VCN and a subnet in which to run the functions. The one guideline here is to pick the subnets that are in the same region as the Docker registry you specified in your context YAML earlier - check these [docs](https://docs.cloud.oracle.com/iaas/Content/Functions/Tasks/functionscreatingapps.htm) for more information.
```
mvandervliet: Are there any included links to install the fn CLI?

src/docs/source/includes/_deployment.md
peterj: It's truncated, there's not much to do, unless you have an idea for a shorter name.

src/docs/source/includes/_deployment.md
peterj: added the section

peterj:
I think we need to use the same docs or remove the READMEs and use gh-pages (or vice-versa). Maintaining the exact same content in both places doesn't make sense.

mvandervliet:
> I think we need to use the same docs or remove the READMEs and use gh-pages (or vice-versa). Maintaining the exact same content in both places doesn't make sense.

Personally I find it useful to have a per-service doc in its respective `src/**/README.md` and the docs for entire functioning app (with backing service setup instructions) in `gh-pages`. As an example, `src/functions/newsletter...` could contain information about build/deploy/test of that function itself, whereas `gh-pages` for the same would present the way to configure this for the application runtime

PR #82, opened 2020-01-31
fix randomly generated schema credentials with leading numeric

mvandervliet:
Fixes problem that arises with a randomly generated password beginning with a numeric character.  The problem cascades into the ATP schema initialization scripts and fails to create the `USER`. Additional steps are necessary to handle this scenario in cases where user-provided password begins with a digit.

### DIFF
--- a/deploy/complete/helm-chart/provision/templates/secrets-global.yaml
+++ b/deploy/complete/helm-chart/provision/templates/secrets-global.yaml
@@ -2,7 +2,7 @@
 {{- $oadbWalletPw := .Values.global.oadbWalletPassword | default (randAlphaNum 12) }}
 {{- $oadbName := .Values.global.oadbName | default (printf "mushop%s" (randAlphaNum 12)) }}
 {{- $oadbUser := .Values.global.oadbUser | default (printf "user%s" (randAlphaNum 12)) }}
-{{- $oadbPw := .Values.global.oadbPassword | default (printf "%s" (randAlphaNum 12)) }}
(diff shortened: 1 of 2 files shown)


deploy/complete/helm-chart/provision/templates/secrets-global.yaml
```diff
@@ -2,7 +2,7 @@
 {{- $oadbWalletPw := .Values.global.oadbWalletPassword | default (randAlphaNum 12) }}
 {{- $oadbName := .Values.global.oadbName | default (printf "mushop%s" (randAlphaNum 12)) }}
 {{- $oadbUser := .Values.global.oadbUser | default (printf "user%s" (randAlphaNum 12)) }}
-{{- $oadbPw := .Values.global.oadbPassword | default (printf "%s" (randAlphaNum 12)) }}
+{{- $oadbPw := .Values.global.oadbPassword | default (printf "%s%s" (randAlpha 1) (randAlphaNum 12)) }}
```
junior: meets the restrictions, can even be more chars than 13:
```
Passwords can be from 1 to 30 characters.

The first character in an Oracle password must be a letter.

Only letters, numbers, and the symbols “#”, “_” and “$” are acceptable in a password.```

junior: Some cultures does not like the number 13, to avoid any references, like the santagate.

There's even religious references: https://en.wikipedia.org/wiki/13_(number)

PR #83, opened 2020-02-01
fix on the objectstorage name and provision values

junior:


deploy/complete/helm-chart/provision/templates/service-instance-global.yaml
```diff
@@ -51,7 +51,7 @@ spec:
   clusterServiceClassExternalName: object-store-service
   clusterServicePlanExternalName: standard
   parameters:
-    name: {{ (printf "MuShop %s Catalogue Bucket" (randAlphaNum 4)) | quote }}
+    name: {{ (printf "MuShop_%s_Catalogue_Bucket" (randAlphaNum 4)) | quote }}
```
mvandervliet: nit: 
Because it's a bucket that's created, there's no need to add "Bucket" to the name. Also, preferred qualifier would be "media" instead of "catalogue"

deploy/complete/helm-chart/provision/values.yaml
```diff
@@ -2,6 +2,12 @@ global:
   osb:
     # Set to true to create a global ATP instance
     atp: true
+    # Set to true to create a streaming service instance
+    oss: true
+    # Set to true to create a bucket on the object storage service instance
+    objectstorage: true
+    # Set object storage namespace, usualy the tenancy name or ocid.
+    objectstoragenamespace: ociateam
```
mvandervliet: shouldn't have our tenancy name here. In the comment, it states `ocid` can be used? Lastly, why not just nest this as:

```yaml
osb:
  objectstorage:
    namespace: mytenancy
```

deploy/complete/helm-chart/provision/templates/service-instance-global.yaml
junior: Just following the original on the DB: `name: {{ (printf "MuShop %s Db" (randAlphaNum 4)) | quote }}`. There's Db everywhere. 
image


Those are not generic media, just serve the catalogue images related to products. There's no web assets.
Propose to change to `MuShop_%s_Catalogue_Media_Images`

deploy/complete/helm-chart/provision/values.yaml
junior: This was misplaced. Following the values-prod-reference.yaml on the mushop chart. For future changes, need to decide if will separate the creation of the bucket during the provision chart and the load of the images using the mushop chart or if will do all together.

deploy/complete/helm-chart/provision/templates/service-instance-global.yaml
junior: or `MuShop_%s_Product_Images`

deploy/complete/helm-chart/provision/values.yaml
mvandervliet: Sounds good. I am almost done with the OOS bucket auto-population, and the separation of the images (assets service) from `catalogue`.

PR #85, opened 2020-02-04
Feature/static asset service

mvandervliet:
Incorporates `assets` service for the following functionality:
- optimize and deploy (job) product/hero images to Object Storage
- lightweight runtime (deployment) for serving images when OOS is not configured (default)
- consume OSB objectstorage binding for OOS `PAR` bucket population

Next steps:
- ~update TF stack to create bucket/PAR for images~
- ~update build for ORM to include asset optimization~
- ~update ORM setup scripts to deploy images to OOS bucket when node is started~
- ~update ORM script to configure object storage URL for assets in OOS~
- remove `images` from `src/catalogue`
- update service diagram


deploy/complete/helm-chart/mushop/charts/assets/values.yaml
```diff
@@ -0,0 +1,40 @@
+replicaCount: 1
+
+image:
+  repository: iad.ocir.io/oracle/ateam/mushop-assets
+  tag: 1.0.0
```
junior: will be using major.minor.patch for this image? will need to update the yaml every time there's a small patch for this service.

deploy/complete/helm-chart/mushop/charts/edge/values.yaml
```diff
@@ -2,7 +2,7 @@ replicaCount: 1
 
 image:
   repository: iad.ocir.io/oracle/ateam/mushop-edge-router
-  tag: 1
+  tag: 1.1.0
```
junior: Should not be 1.1?

deploy/complete/helm-chart/mushop/charts/assets/values.yaml
mvandervliet: I've found it preferable this way during development. Also when giving the workshops, to have `imagePullPolicy: IfNotPresent`. I agree it's nice to avoid frequent edits to the yaml, but the higher degree of consistency seems worth it

deploy/complete/helm-chart/mushop/charts/edge/values.yaml
mvandervliet: Same thought as above, mostly to be consistent

deploy/complete/helm-chart/mushop/charts/assets/values.yaml
junior: The question come up because this was discoursed and set in previous meetings to sue major.minor or just major depending on the case. Should be the same tagging strategy for all services.

mvandervliet: I remember the discussions well, however there was never absolute consensus on the tags to use in the helm charts. The use of fuzzy `<major>.<minor>` with pull `Always` is only done in the `edge` and `catalogue` services today - the rest use full semver with `IfNotPresent`. In developing certain services I like the fuzzy version, but semver for others. Realistically though, we want the production runtime to reflect any change, even patch versions. To do so properly, we must update the deployments anyway.

deploy/basic/Dockerfile
```diff
@@ -121,16 +151,15 @@ FROM base
...
 
 # Copy Services apps
 COPY --from=storefront-builder /app/storefront/build /app/storefront
 COPY --from=api-builder /app/api /app/api
+COPY --from=assets-builder /app/assets /app/assets
 COPY --from=catalogue-builder /catalogue /app/catalogue/catalogue
 
 # Create zip package of the Apps and local images
-RUN mkdir /package && GZIP=-9 tar cvzf /package/mushop-basic.tar.gz /app /images
+RUN mkdir /package && GZIP=-9 tar cvzf /package/mushop-basic.tar.gz /app
```
junior: with the new changes, what is the new size of the stack.zip?

deploy/basic/entrypoint.sh
```diff
@@ -12,19 +12,10 @@
...
 export NODE_ENV=production
 export CATALOGUE_PORT=3005
-export CATALOGUE_URL=http://localhost:3005
-export USERS_URL=http://user
+export CATALOGUE_URL=http://localhost:${CATALOGUE_PORT}
 
 echo "Environment: $(uname -a)";
 
-echo "Checking Installation..."
-if [ ! -d "/app/api/node_modules" ]
```
junior: the node_modules is now included on the stack.zip?

deploy/basic/terraform/storage.tf
```diff
@@ -128,4 +128,24 @@ resource "oci_objectstorage_object_lifecycle_policy" "mushop_deploy_assets_lifec
...
+}
+
+# Static assets PAR
+resource "oci_objectstorage_preauthrequest" "mushop_media_preauth" {
+  #Required
+  access_type  = "AnyObjectWrite"
+  bucket       = oci_objectstorage_bucket.mushop_media.name
+  name         = "mushop_assets_preauth"
+  namespace    = data.oci_objectstorage_namespace.user_namespace.namespace
+  time_expires = timeadd(timestamp(), "72h")
```
junior: The par expiration can be reduced to match the others "to be expired" objects. This PAR is used just during the installation.

deploy/basic/entrypoint.sh
mvandervliet: no. The install is moved to `node.sh`

deploy/basic/Dockerfile
mvandervliet: 7.8Mb

junior:
If these changes results in a new release of the always-free stack, please tag the release.

mvandervliet:
> If these changes results in a new release of the always-free stack, please tag the release.

I will, once merged (with update to readme as well)

PR #86, opened 2020-02-05
Fix style broken when using <aside call=

junior:
Move warning bellow ```shell entry


junior:
Using `<aside class="warning">...</aside>`  before ` ```shell--helm` was breaking the style.

PR #89, opened 2020-02-06
unit tests for catalogue service

junior:
Basic Unit tests for service.go. 80% Coverage.

### DIFF
--- a/src/catalogue/go.mod
+++ b/src/catalogue/go.mod
@@ -3,6 +3,7 @@ module mushop/catalogue
 go 1.13
 
 require (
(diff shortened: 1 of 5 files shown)


src/catalogue/go.mod
```diff
@@ -3,6 +3,7 @@ module mushop/catalogue
 go 1.13
 
 require (
+	github.com/DATA-DOG/go-sqlmock v1.4.1
```
mvandervliet: Is this something that could also be used for “mock” mode? Thereby allowing the API to use all actual micro services, with mocked backing services?

junior: Yes, I'm planning something similar.

PR #90, opened 2020-02-06
TF Assets destroy

mvandervliet:
Assets now explicitly added to the tf stack. This resolves `BucketNotEmpty` error when destroying stack.

Minor fixes to helm chart provisioning

### DIFF
--- a/deploy/complete/helm-chart/provision/templates/cluster-service-broker.yaml
+++ b/deploy/complete/helm-chart/provision/templates/cluster-service-broker.yaml
@@ -2,7 +2,7 @@
 apiVersion: servicecatalog.k8s.io/v1beta1
 kind: ClusterServiceBroker
 metadata:
-  name: {{.Values.global.osb.name}}
+  name: oci-cluster-service-broker
   annotations:
     "helm.sh/resource-policy": keep
   labels:
--- a/deploy/complete/helm-chart/provision/templates/service-binding-global.yaml
+++ b/deploy/complete/helm-chart/provision/templates/service-binding-global.yaml
@@ -36,4 +36,6 @@ metadata:
 spec:
   instanceRef:
     name: {{ .Values.global.osb.instanceName }}-objectstorage
+  parameters:
+    generatePreAuth: true
 {{ end }}
--- a/deploy/complete/helm-chart/provision/values.yaml
+++ b/deploy/complete/helm-chart/provision/values.yaml
@@ -9,7 +9,7 @@ global:
     # Set object storage namespace, usualy the tenancy name or ocid.
     objectstoragenamespace: ocitenancyname
     # Service broker name (used in the service URL)
-    name: mushop-osb
+    name: oci-broker
     port: 8080
     compartmentId:
     # Service instance name
--- a/src/docs/source/includes/_deployment.md
+++ b/src/docs/source/includes/_deployment.md
@@ -262,8 +262,6 @@ secret to the `mushop-utilities` namespace:
       --set global.osb.compartmentId=<compartmentId>
     ```
 
-    > Note that the `oci-credentials` secret was created [previously](#provisioning)
-
 1. It will take a few minutes for the ATP database to provision, and the Wallet binding to become available. Verify `serviceinstances` and `servicebindings` are **READY**:
 
     ```text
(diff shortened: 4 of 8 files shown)


deploy/complete/helm-chart/provision/templates/service-binding-global.yaml
```diff
@@ -36,4 +36,6 @@ metadata:
 spec:
   instanceRef:
     name: {{ .Values.global.osb.instanceName }}-objectstorage
+  parameters:
+    generatePreAuth: true
```
mvandervliet: @junior  this is the reason for the empty binding. Note too that we have an issue where the `serviceinstance` cannot be deleted due to objects in the bucket. Let's discuss a solution for that

deploy/complete/helm-chart/provision/templates/cluster-service-broker.yaml
```diff
@@ -2,7 +2,7 @@
 apiVersion: servicecatalog.k8s.io/v1beta1
 kind: ClusterServiceBroker
 metadata:
-  name: {{.Values.global.osb.name}}
+  name: oci-cluster-service-broker
```
mvandervliet: differentiate the oci-broker from the cluster service broker resource

deploy/complete/helm-chart/provision/values.yaml
```diff
@@ -9,7 +9,7 @@ global:
     # Set object storage namespace, usualy the tenancy name or ocid.
     objectstoragenamespace: ocitenancyname
     # Service broker name (used in the service URL)
-    name: mushop-osb
+    name: oci-broker
```
mvandervliet: Change required otherwise the instructions fail.

deploy/complete/helm-chart/provision/templates/service-binding-global.yaml
junior: The TF should track each object inside the bucket to safely delete the objects with TF Destroy. This affects the Always-Free Terraform scripts. The Binding be created by Helm is different, on the Complete Terraform Scripts, which is handled by the terraform helm provider.

deploy/complete/helm-chart/provision/templates/cluster-service-broker.yaml
junior: image

deploy/complete/helm-chart/provision/templates/service-binding-global.yaml
junior: Include helm changes on the basic TF PR confused a lit bit. Why cannot delete objects from TF?

deploy/complete/helm-chart/provision/values.yaml
junior: ok. missed on the previous PR.

src/docs/source/includes/_deployment.md
```diff
@@ -262,8 +262,6 @@ secret to the `mushop-utilities` namespace:
       --set global.osb.compartmentId=<compartmentId>
     ```
 
-    > Note that the `oci-credentials` secret was created [previously](#provisioning)
```
junior: We need a note saying where the oci-credentials is coming from.

deploy/complete/helm-chart/provision/templates/service-binding-global.yaml
mvandervliet: The TF stack now does track each bucket object in the config. The helm chart contained no PAR for image asset deployment, hence this change. But, once images are added (from K8s, or otherwise manually, or through an admin interface) then the bucket cannot be deleted.

src/docs/source/includes/_deployment.md
mvandervliet: It is there in step 1 for OSB

mvandervliet:
We’ll need to address the problem of bucket de provisioning. OSB fails in this regard. OOS needs to support a `force_delete` option like S3

deploy/complete/helm-chart/provision/templates/service-binding-global.yaml
junior: If the images are tracked within the helm provider to the tf state, will be deleted. 

Will not be deleted if new products with new images (or images changes for original products) be included. This is valid for both, basic and complete.

junior:
> We’ll need to address the problem of bucket de provisioning. OSB fails in this regard. OOS needs to support a `force_delete` option like S3

Agree. There's a option to do "force", but still do not delete bucket with contents.

PR #109, opened 2020-02-18
Events/tracking service addition

mvandervliet:
Complete:

- Service dev (go)
- Docker support
- K8s/helmify
- Implemented for api


src/events/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+| Variable | Description | Default | Required |
+|---|---|---|---|
+| `TENANCY` | Tenancy OCID string |  | Y |
+| `USER_ID` | IAM User OCID string |  | Y |
+| `PRIVATE_KEY` | API Private Key (pem) |  | Y |
+| `FINGERPRINT` | API Key Fingerprint  |  | Y |
+| `PASSPHRASE` | API Key passphrase  |  | N |
+| `REGION` | Oracle Cloud Infrastructure region (ex: `us-ashburn-1`)  |  | Y |
+| `STREAM_ID` | Stream OCID string |  | Y |
```
junior: when set the k8s secret to consume stream_id and message_endpoint, would be better to be separated from the other variables, to help the common usage of the oci service broker.

mvandervliet: > when set the k8s secret to consume stream_id and message_endpoint, would be better to be separated from the other variables, to help the common usage of the oci service broker.

Yes, that's the plan: `oci-credentials` + `oss-binding`.  Once OSB supports the streamPoolId, then it can just use kafka compat with secrets: OSB binding + auth token

src/events/api-spec/hooks.js
```diff
@@ -0,0 +1,22 @@
+/**
+ * Copyright © 2020, Oracle and/or its affiliates. All rights reserved.
+ * The Universal Permissive License (UPL), Version 1.0
+ */
+const hooks = require('hooks');
+
+// Mock connection
+hooks.before("/events > POST", (transaction, done) => {
```
peterj: what is the purpose of this file?

src/events/cmd/main.go
```diff
@@ -0,0 +1,104 @@
...
+)
+
+const (
+	ServiceName = "events"
+)
+
+func main() {
+	var (
+		port = flag.String("port", "8080", "Port to bind HTTP listener")
+		zipk = flag.String("zipkin", os.Getenv("ZIPKIN"), "Zipkin address")
```
peterj: we don't use this, right?

src/events/config.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"os"
+
+	"github.com/oracle/oci-go-sdk/common"
+	"github.com/oracle/oci-go-sdk/streaming"
+)
+
+// EnvironmentConfigurationProvider uses environment variables to get OCI config
+func EnvironmentConfigurationProvider() (common.ConfigurationProvider, error) {
+	pass := os.Getenv("PASSPHRASE")
```
peterj: any reason for `pass` being outside? (is it because it can be empty?)

src/events/config.go
```diff
@@ -0,0 +1,44 @@
...
+	)
+	_, err := common.IsConfigurationProviderValid(provider)
+
+	return provider, err
+}
+
+// GetStreamClient returns a streaming client with the given configuration provider
+func GetStreamClient(provider common.ConfigurationProvider) (streaming.StreamClient, error) {
+	var endpoint string
+	endpoint = os.Getenv("MESSAGES_ENDPOINT")
```
peterj: could this be just `endpoint := os.Getenv("MESSAGES_ENDPOINT")` instead of explicitly declaring it on line 32?

src/events/endpoints.go
```diff
@@ -0,0 +1,73 @@
...
+import (
+	"github.com/go-kit/kit/endpoint"
+	"github.com/go-kit/kit/tracing/opentracing"
+	stdopentracing "github.com/opentracing/opentracing-go"
+	"golang.org/x/net/context"
+)
+
+// Endpoints collects the endpoints that comprise the Service.
+type Endpoints struct {
+	EventsEndpoint endpoint.Endpoint
```
peterj: nit: these should probably just be named `Events` and `Health` as they already live inside the Endpoints struct (i.e. no need to repeat the Endpoint again)

src/events/endpoints.go
```diff
@@ -0,0 +1,73 @@
...
+		span, ctx = stdopentracing.StartSpanFromContext(ctx, "health check")
+		span.SetTag("service", "events")
+		defer span.Finish()
+		health := s.Health()
+		return healthResponse{Health: health}, nil
+	}
+}
+
+// EventsRequest represents a request for events tracking
+type EventsRequest struct {
```
peterj: nit: similar as above -- the package is named `events`, so having the name `Events` repeat in structs is odd. (e.g. `events.EventsResponse` or `events.EventsRequest`). It should be `events.Response` and `events.Request`

src/events/logging.go
```diff
@@ -0,0 +1,50 @@
...
+		}
+	}
+}
+
+type loggingMiddleware struct {
+	next   Service
+	logger log.Logger
+}
+
+func (mw loggingMiddleware) EventsReceiver(source string, track string, events []Event) (received EventsReceived, err error) {
```
peterj: nit: `Receiver`

src/events/service.go
```diff
@@ -0,0 +1,126 @@
...
+	s.logger.Log(
+		"source", source,
+		"track", track,
+		"length", numEvents,
+	)
+
+	var err error
+	received := false
+
+	if numEvents > 0 {
```
peterj: I'd invert this if (`if numEvents <= 0`) and get rid of the big `if` block and an extra level of indentation :)

src/events/service.go
```diff
@@ -0,0 +1,126 @@
...
+
+		// construct messages
+		var messages []streaming.PutMessagesDetailsEntry
+
+		for _, evt := range events {
+			msg := EventRecord{
```
peterj: why not this:

```
msg := EventRecord {
    Source: source,
    Track: track,
    Type: evt.Type,
    Detail: evt.Detail,
}
```

src/events/api-spec/hooks.js
mvandervliet: Defines a mock invocation of the service endpoint

src/events/cmd/main.go
mvandervliet: not yet, but is consistent with the other `go` services. So would be simple to add/configure zipkin if desired

src/events/config.go
mvandervliet: because the function signature requires a pointer: https://github.com/oracle/oci-go-sdk/blob/master/common/configuration.go#L55

src/events/config.go
mvandervliet: It could, but then it would always require 2️⃣ properties, instead of just the 1️⃣  `STREAM_ID`. Essentially it's a fallback logic in cases where OSB is not used, and a simpler `oss-connection` secret definition is desired.

src/events/service.go
mvandervliet: will have a look, but would likely use `== 0` instead since `< 0` isn't possible

src/events/service.go
mvandervliet: 🤷‍♂ error: `cannot use promoted field Event.Time in struct literal of type EventRecord`

src/events/endpoints.go
mvandervliet: I think this should be the left as-is, since there is more than one `Request` type in the service.

mvandervliet:
Ready for final review

src/api/api/common/index.js
```diff
@@ -25,4 +25,11 @@ module.exports = {
     return await axios.get(`${endpoints.cartsUrl}/${id}/items`)
       .then(({ data }) => data)
   },
+
+  getSvcTracker: source => {
+    return (req, type, detail = {}) =>
+      helpers.trackEvents(req, { source, events: [{type, detail}] })
+        .catch(() => { /* noop */ });
```
mvandervliet: Graceful event failures

mvandervliet: > What is the behavior if the Streaming Service is not provisioned? The tracker get ignored?

Essentially, yes. The `events` pod can enter an error state (considered normal), whereas all attempts to track from the api will fail gracefully

jeevanjoseph:
Out of curiosity  - Instead of peppering the various services with calls to tracker that posts the tracking data, would an [axios interceptor](https://github.com/axios/axios#interceptors] have any merit ? Of course, the interceptor should not try to track the tracking calls. This might have the benefit of decorating the event data with service call outcome as well. Not sure if we would lose the expressiveness/readability of the tracker middleware.

mvandervliet:
> Out of curiosity - Instead of peppering the various services with calls to tracker that posts the tracking data, would an [axios interceptor](https://github.com/axios/axios#interceptors) have any merit ? Of course, the interceptor should not try to track the tracking calls. I'm also not sure if axios would prevent an http call from within an interceptor. This might have the benefit of decorating the event data with service call outcome as well. Not sure if we would lose the expressiveness/readability of the tracker middleware though.

I like this idea a lot, however, it would seem to resemble more of a tracing feature. There is also the challenge of interceptor affinity to the requestor (client uuid), which (at least with axios) is not so straightforward. I'll do some more thinking on this 🤔

PR #115, opened 2020-02-19
Build & push docker images on VERSION file changes

peterj:


.github/workflows/dockerpush.yml
```diff
@@ -0,0 +1,44 @@
+name: Build and push Docker images
+
+on:
+  push:
+    paths:
+      # Match any VERSION file under /src folder
+      - 'src/**/VERSION'
+env:
+  DOCKER_REGISTRY_URL: iad.ocir.io
+  REPOSITORY: oracle/mushop
```
junior: This may be more convenient if be a secret, so we can change the variable without change the workflow file.

peterj:
true; I had it as a secret originally. The only downside is that you don’t
see the full name in the logs as it will be redacted. I guess we can live
with that.

PR #120, opened 2020-02-21
Update the workflow to push all image versions

peterj:


peterj:
I logged #121 for that. We can see how complicated would it be to support multiple lines and if it would be worth implementing it (assuming that the build-arg option is not something we will do all the time and it perhaps might be worth doing it as a one-off when needed)

PR #126, opened 2020-02-26
doc po watch

mvandervliet:


deploy/complete/helm-chart/mushop/values-dev.yaml
```diff
@@ -6,6 +6,10 @@ global:
   oadbConnectionSecret: oadb-connection # Name of connection secret created separately
   oosBucketSecret: oos-bucket           # Object storage bucket secret name (optional)
 
+ingress:
+  hosts: # Configure ingress to support host-specific routing
+    # - yourname.mushop.com
```
junior: What do you think to leave a uncommented placeholder?
```suggestion
     - mushop.local #  yourmushopdomain.com 
```

mvandervliet: if we do this, then default ingress will *only* match the `mushop.local` dns usage

junior: unless the deployer user change it to something else. And can enable the IP address as well.

PR #136, opened 2020-03-09
Add Istio docs

peterj:


src/docs/source/includes/_deployment.md
```diff
@@ -503,8 +503,10 @@ helm chart is installed using settings to leverage cloud backing services.
     mushop-user-init-1-8k62c              0/1     Completed   0          33s
     ```
 
-1. Open a browser with the `EXTERNAL-IP` created during setup, **OR** `port-forward`
-directly to the `edge` service resource:
+>Note: if you installed Istio service mesh, you should see `2/2` in the `READY` column and `1/2` for the init pods and the assets deploy pod.
```
junior: Do you think that makes sense to explain why you see 2/2? Like saying that is because of sidecar...

src/docs/source/includes/_deployment.md
```diff
@@ -520,4 +522,13 @@ directly to the `edge` service resource:
     ```
 
     > Locating `EXTERNAL-IP` for Ingress Controller. **NOTE** this will be
-    [localhost](https://localhost) on local clusters.
\ No newline at end of file
+    [localhost](https://localhost) on local clusters.
```
junior: There's a clear tag saying "Istio" for that option?

src/docs/source/includes/_deployment.md
peterj: I mention sidecars in the Istio portion, but I can add it here as well if that makes sense

src/docs/source/includes/_deployment.md
peterj: what do you mean?

junior: Just asking if there’s a separation of “with isitio” and “without sítio”. I saw that there’s a note “if you are using istio”. 

I didn’t rendered the docs yet, do not know how is showing.

src/docs/source/includes/topics/_istio.md
```diff
@@ -0,0 +1,189 @@
...
+kubectl create ns mushop
+```
+
+1. Label the namespace with `istio-injection=enabled`:
+
+```shell
+kubectl label namespace mushop istio-injection=enabled
+```
+
+1. Follow the instructions for <a href="#deployment">deploying Mushop</a>.
```
mvandervliet: use markdown for link: `[deploying MuShop](#deployment)`

src/docs/source/includes/topics/_istio.md
```diff
@@ -0,0 +1,189 @@
...
+  servers:
+    - port:
+        number: 80
+        name: http
+        protocol: HTTP
+      hosts:
+        - '*'
+EOF
+```
+
```
mvandervliet: Can you create windows versions of these commands as well? something like how they do for AKS https://docs.microsoft.com/en-us/azure/aks/servicemesh-istio-install?pivots=client-operating-system-windows#install-the-istio-components-on-aks

src/docs/source/includes/_deployment.md
peterj: Istio is under a separate heading, however, since I am referring back to the deployment I figured to add the "If using Istio" just in case.

PR #145, opened 2020-03-16
Feature: healthy events

mvandervliet:
- Events container will appear healthy even when misconfigured
- Service requests (`POST /events`) will respond with 500 in this state


jeevanjoseph:
What’s the difference between inadvertently misconfiguring the stream vs mocking it ?

mvandervliet:
> What’s the difference between inadvertently misconfiguring the stream vs mocking it ?

None presently. However logs would indicate the problem. Example:

```
ts=2020-03-16T22:47:22.551449Z caller=main.go:86 error="tenancy OCID can not be empty"
ts=2020-03-16T22:47:22.551538Z caller=main.go:87 warning="Running without stream configuration"
```

Or would you like some specific logic around passing `"mock"` as some env with alternate behavior (clearly more involved)

mvandervliet:
> This PR also solves the error state of the pod if there's no Stream service?

yes

jeevanjoseph:
explicitly setting something to “ignore” stream connections would be nice but no need to block on it. The error message should be sufficient to troubleshoot if needed.

PR #149, opened 2020-04-24
Updating SODA to 1.1.4 and JDBC to 19.6

jjspiegel:
- Updates JDBC to 19.6 and uses new oracle.database coordinates
- Updates SODA to 1.1.4
- Adds metadata caching optimization

Note, ATP and ADW switches over to binary JSON storage on 4/28.  This update should be tested and deployed in advance of that.  The old JDBC/SODA jars will not work against binary JSON storage.


src/carts/src/main/java/mushop/carts/CartRepositoryDatabaseImpl.java
```diff
@@ -156,4 +160,12 @@ public boolean healthCheck() {
         }
     }
 
+    static {
+        // Optimization: cache collection metadata to avoid extra roundtrips
+        // to the database when opening a collection
+        Properties props = new Properties();
+        props.put("oracle.soda.sharedMetadataCache", "true");
+        props.put("oracle.soda.localMetadataCache", "true");
```
morgiyan, line 168: I'd remove local cahce (line 168).

jjspiegel: Done.

src/carts/Dockerfile
```diff
@@ -1,7 +1,7 @@
 ##### Application Metadata
 ARG APPLICATION_NAME="carts"
 ARG VERSION="1.1.0-SNAPSHOT"
-ARG SODA_VERSION="1.1.0"
+ARG SODA_VERSION="1.1.4"
 ARG clientVersion=19.3
```
junior, line 5: Just a note. Even using 19.6, this argument need to remain 19.3 for compatibility with the volume created. There's a ticket open to improve how we get the wallets, and them this will not be needed.

src/carts/README.md
```diff
@@ -70,7 +70,7 @@ export PATH=$JAVA_HOME/bin:/path/to/maven/bin
 To compile the service:
 
 ```bash
-mvn clean package
+mvn clean package -DskipTests
```
jeevanjoseph, line 73: Why skip tests ?

jjspiegel: The unit tests are currently running on port 80 which is a privileged port so running locally will, for many people, raise an error.  They still run in the docker build.  The point of these instructions is to just let the service run standalone so running the tests isn't critical.  The alternative is to allow the port to be configured to something over 1024 (which may be possible already).  I just don't have the cycles to look into it at the moment.

PR #169, opened 2020-06-09
Feature/ojdbc client updates

jeevanjoseph:
Database connection details (wallet contents) are mounted at `/wallet` now. These applications no longer use the Oracle instant client since the JDBC libs are available on maven central now. The old path (`/usr/lib/oracle/19.3/client64/lib/network/admin/`) was to be compliant with the instant client - which is no longer required. 

As a side effect, now the jdbc driver versions are controlled directly in the `build.gradle` like any other application dependency. No more version numbers in `Dockerfile` and helm charts. They all consistently point to `/wallet`.

Upgrade to JDBC driver version `19.6.0.0`


junior:
Missing change the VERSION file to generate the image. And reference of the version on the Kubernetes Manifest/Helm.

src/orders/Dockerfile
```diff
@@ -63,7 +61,8 @@ EXPOSE 80
 WORKDIR /app
 ENV APPLICATION_NAME=${APPLICATION_NAME}
 ENV VERSION=${VERSION}
-ENV TNS_ADMIN=/usr/lib/oracle/${clientVersion}/client64/lib/network/admin/
+# Edit the helm chart to map
```
junior, line 64: "Edit the helm chart to map" ?

jeevanjoseph: This is just a comment to users who want to dig a bit deeper to map the wallets themselves etc. Is the comment that this is too vague ?(I agree its valgue) . Shall I make it a bit more explicit and spell it out ?

junior: Remove comment and put a note somewhere. It's not common for users to mess around wallets. I cannot even see a use case.

jeevanjoseph: removed.

Issue #163, opened 2020-06-03
Error while trying to run Mushop docs locally


naikvenu:
While trying to run docs locally i get this error:

```
$pwd
oci-cloudnative/src/docs
$ make dev
cd themes/redwood && npm install

> mushop-docs-theme@1.0.0 postinstall /Users/vsnaik/Desktop/incubation-team/mushop-github/oci-cloudnative/src/docs/themes/redwood
> cp -r node_modules assets/js/

npm WARN mushop-docs-theme@1.0.0 No description
npm WARN mushop-docs-theme@1.0.0 No repository field.

added 75 packages from 69 contributors and audited 75 packages in 4.043s
found 0 vulnerabilities



   ╭────────────────────────────────────────────────────────────────╮
   │                                                                │
   │      New patch version of npm available! 6.14.4 → 6.14.5       │
   │   Changelog: https://github.com/npm/cli/releases/tag/v6.14.5   │
   │               Run npm install -g npm to update!                │
   │                                                                │
   ╰────────────────────────────────────────────────────────────────╯

Building sites … ERROR 2020/06/03 20:28:32 render of "page" failed: execute of template failed: template: _default/single.html:1:3: executing "_default/single.html" at <partial "_open.html" .>: error calling partial: execute of template failed: template: partials/_open.html:5:7: executing "partials/_open.html" at <partial "menu.html" .>: error calling partial: "/Users/vsnaik/Desktop/incubation-team/mushop-github/oci-cloudnative/src/docs/themes/redwood/layouts/partials/menu.html:93:24": execute of template failed: template: partials/menu.html:93:24: executing "tree-nav" at <(not .Params.hidden) or $.showhidden>: can't give argument to non-function not .Params.hidden
ERROR 2020/06/03 20:28:32 render of "page" failed: execute of template failed: template: _default/single.html:1:3: executing "_default/single.html" at <partial "_open.html" .>: error calling partial: execute of template failed: template: partials/_open.html:5:7: executing "partials/_open.html" at <partial "menu.html" .>: error calling partial: "/Users/vsnaik/Desktop/incubation-team/mushop-github/oci-cloudnative/src/docs/themes/redwood/layouts/partials/menu.html:93:24": execute of template failed: template: partials/menu.html:93:24: executing "tree-nav" at <(not .Params.hidden) or $.showhidden>: can't give argument to non-function not .Params.hidden
[...]

PR #174, opened 2020-06-23
Feature/setup buildtools

jeevanjoseph:
- adds a Jenkins build system to the setup charts
- configured to use the Kubernetes plugin to manage build jobs
- disabled by default, but can be enabled by setting `jenkins.enabled`
eg: 
```
  helm install  mushop-utils deploy/complete/helm-chart/setup \
  --namespace mushop-utilities \
  --set jenkins.enabled=true 
```

- ingress is used by default, but can be disabled by setting `jenkins.master.ingress.enabled`
eg: 
```
helm install  mushop-utils deploy/complete/helm-chart/setup \
  --namespace mushop-utilities \
  --set jenkins.enabled=true \
  --set jenkins.master.ingress.enabled=false
```


deploy/complete/helm-chart/setup/templates/NOTES.txt
```diff
@@ -44,6 +44,36 @@ resource in this namespace.
...
+
+An instance of Jenkins was installed with this chart. These instructions describe
+how to access the Jenkins application. 
+
+- Get the admin password:
+
+  kubectl get secret -n {{ .Release.Namespace }} {{ .Release.Name }}-jenkins \
+    -o jsonpath="{.data.jenkins-admin-password}" | base64 --decode ; echo
+
+- Ensure that jenkins is up and running
```
peterj, line 59: nit: Jenkins

deploy/complete/helm-chart/setup/values.yaml
```diff
@@ -76,4 +76,47 @@ grafana:
     - grafana-kubernetes-app
     - grafana-worldmap-panel
     - grafana-piechart-panel
-    - btplc-status-dot-panel
\ No newline at end of file
+    - btplc-status-dot-panel
+
+# https://github.com/helm/charts/blob/master/stable/jenkins/README.md
+jenkins:
```
mvandervliet, line 82: How much of these are defaults? Is it possible to simplify our values by deferring to the jenkins chart defaults?

jeevanjoseph: Quite a few - good catch.
eliminated most of it. The one thing I'd like to keep as a redundant one is the `serviceType=ClusterIP`. This is default, but that could change, since the default for the rest of the config is `NodePort`.

jeevanjoseph: @mvandervliet - having some issues running `hugo`. Will add docs as soon as that is resolved.

```
Building sites … ERROR 2020/06/23 23:57:06 render of "page" failed: execute of template failed: template: _default/single.html:1:3: executing "_default/single.html" at <partial "_open.html" .>: error calling partial: execute of template failed: template: partials/_open.html:5:7: executing "partials/_open.html" at <partial "menu.html" .>: error calling partial: "/Users/JGJOSEPH/dev/mushop/src/docs/themes/redwood/layouts/partials/menu.html:93:24": execute of template failed: template: partials/menu.html:93:24: executing "tree-nav" at <(not .Params.hidden) or $.showhidden>: can't give argument to non-function not .Params.hidden
```

mvandervliet: ok... this probably relates to #163, I'll take a look

deploy/complete/helm-chart/setup/templates/NOTES.txt
jeevanjoseph: Fixed

deploy/complete/helm-chart/setup/values.yaml
jeevanjoseph: @mvandervliet - thanks for the help with `hugo`. Docs updated - the "how to setup a pipeline" section is scant right now. 
Will expand later as a HOL

PR #179, opened 2020-07-04
Feature/nats metrics

jeevanjoseph:
- Enables metrics collection for NATS server.
- adds the Prometheus exporter as an additional container in the NATS pod. 
- the exporter exposes the `/metrics` endpoint that exposes the metrics in prometheus exposition format.


deploy/complete/helm-chart/mushop/charts/nats/values.yaml
```diff
@@ -5,6 +5,11 @@ image:
   tag: 2.1.2
   pullPolicy: IfNotPresent
 
+exporter: 
+  repository: synadia/prometheus-nats-exporter
```
mvandervliet, line 9: Is this the preferred way/image to export nats metrics?

jeevanjoseph: Yes it is. [Docs](https://github.com/nats-io/prometheus-nats-exporter#run) use the same image as well.
Synadia is the "commercial" face of the folks who build NATS.io. 
It simply runs the `prometheus-nats-exporter`. 

We could replace this with our own `Dockerfile` or use the [bitnami image](https://hub.docker.com/r/bitnami/nats-exporter) which is Apache 2.0 licensed.

deploy/complete/helm-chart/mushop/charts/nats/templates/nats-deployment.yaml
```diff
@@ -22,6 +22,9 @@ spec:
         run: {{ include "nats.name" . }}
       annotations:
         sidecar.istio.io/rewriteAppHTTPProbers: "true"
+        prometheus.io/scrape: "true"
```
junior, line 25: scrape or exporter?

jeevanjoseph: scrape - `prometheus-nats-exporter` simply provides a scrape endpoint. The helm chart version can configure prom using the prom operator, in any case, the mechanism is always scrape.

PR #181, opened 2020-07-10
Always free (mushop-basic) stack update

junior:


jeevanjoseph:
Looks good !
We could potentially move a couple of `variables` like `instance_shape` and `autonomous_database_is_free_tier` as visible varibales in `schema.yaml`to let users apply the stack easily on non-free shapes as well. This should also likely change the logic for AD selection.

junior:
> Looks good !
> We could potentially move a couple of `variables` like `instance_shape` and `autonomous_database_is_free_tier` as visible varibales in `schema.yaml`to let users apply the stack easily on non-free shapes as well. This should also likely change the logic for AD selection.

The idea to move everything to variables was exactly make available as "Advanced Options" for the schema. Coming with v1.2.0. Similar that we have with the complete stack.

Issue #138, opened 2020-03-10
Helm chart and service hardcode Oracle DB client version. Should be dynamic.


junior:
Currently Helm chart and service hardcode Oracle DB client version.

This affects the wallet location, the volume for the wallet and the bin folder where the tools are located.

We should either use a variable to select the version and update the dbtools image to use version tag matching the instantClient or create symlinks that will be generic to the chart, no matter which version of the instant client is used.

Reminding that greater version of the client is backwards compatible with "old" DB versions, so, there's no need to have multiple versions support.

PR #185, opened 2020-07-13
Bugfix/orders db init failure

jeevanjoseph:
This is a bug for the orders DB initialization failing when manually setting up ATP. 

The root cause of the issue is that #169 changed the way TNS configuration was done. The goal was to remove the dependency on the instant client version from everywhere except application build (build.gradle/pom.xml). However the change did not account for the fact that the the DB init job uses the instant client, which still expects the old configuration.

This PR effectively creates a `mount` for the same wallet secret for the DB tools container, but mounted on the path that the location it expects. Alternatively it should be sufficient to set the `TNS_ADMIN` env var for the `dbtools` container, however, it does not seem to respect an alternate `TNS_ADMIN`


deploy/complete/helm-chart/mushop/charts/orders/templates/_helpers.tpl
```diff
@@ -94,6 +94,13 @@ app.kubernetes.io/managed-by: {{ .Release.Service }}
   readOnly: true
 {{- end -}}
 
+{{- define "orders.mount.init.wallet" -}}
+# for init container
+- name: wallet
+  mountPath: /usr/lib/oracle/19.3/client64/lib/network/admin/
```
junior, line 100: This is a workaround. We should review to fix the issue #138

jeevanjoseph: Yes - absolutely.

PR #196, opened 2020-08-18
Feature/orders metrics

jeevanjoseph:
Adds micrometer metrics to Orders application. The following metrics are captured now.
- Counters
  - Orders created
  - Orders rejected with rejection reason as a tag/dimension
  - Orders sent to fulfilment
  -  Orders acks received from fulfillment
- Histogram/Distrribution summary
  - Order value (distribution of Order values.) 
  - Note that the standard load test does not generate orders with values that follow as normal distribution, so do not expect to see this plot as a normal  curve when using the standard load test.

Lazy initialization of the spring beans are also enabled for faster startup.

Accompanying grafana dashboard available at https://grafana.com/grafana/dashboards/12696
Dashboard ID `12696`


src/orders/src/main/java/mushop/orders/services/OrdersService.java
```diff
@@ -1,5 +1,7 @@
 package mushop.orders.services;
 
+import io.micrometer.core.instrument.DistributionSummary;
```
mvandervliet, line 3: I'm assuming these don't require any new additions in gradle?

jeevanjoseph: thats correct. Micrometer is part (implicit dependency) of [SpringBoot 2.3 ](https://spring.io/blog/2020/05/15/spring-boot-2-3-0-available-now)- we upgraded to  SpringBoot 2.3 a while back, which automatically gives us access to micrometer.

src/orders/src/main/java/mushop/orders/services/OrdersService.java
```diff
@@ -102,12 +110,28 @@ public CustomerOrder createNewOrder(NewOrderResource orderPayload) {
 
             CustomerOrder savedOrder = customerOrderRepository.save(order);
             LOG.debug("Saved order: " + savedOrder);
+            meterRegistry.summary("orders.amount").record(amount);
+            DistributionSummary.builder("order.stats")
+                    .serviceLevelObjectives(10d,20d,30d,40d,50d,60d,70d,80d,90d,100d,110d)
```
junior, line 115: on future updates, I suggest to use variables instead of literals.

PR #197, opened 2020-08-18
Bugfix/better mock setup

jeevanjoseph:
Updated `values-mock.yaml` to run in mock mode. All relevenat services are addresses.
Additionally, if required, the setup chart can be run to avoid observability and other services (except  the ingress controller).

The command :
```
helm install mushop \
deploy/complete/helm-chart/mushop \
-f deploy/complete/helm-chart/mushop/values-mock.yaml
```
will install the  application with mock mode enabled and no streaming or database related services(or their containers).


jeevanjoseph:
> I think we can improve upon this even more, but for now LGTM

Yes - The tags are not really consistent right now, but a fully tag driven approach might be simpler to use and more flexible.

deploy/complete/helm-chart/mushop/values-mock.yaml
```diff
@@ -2,18 +2,18 @@ global:
...
+  enabled: true
 
-payment:
+nats:
   enabled: false
 
-user:
-  enabled: false
+fulfillment:
+  enabled: false
```
junior, line 19: Please include the new line at the end of the yaml

deploy/complete/helm-chart/mushop/values-mock.yaml
```diff
@@ -2,18 +2,18 @@ global:
...
     service: "all"
 
-# Catalogue service runs in mock mode for product images
-catalogue:
-  enabled: true
+tags:
+  streaming: false
+  atp: false
 
-carts:
```
junior, line 9: Removing the carts specifics make the carts to the deployed, but because ATP = false will make the carts and orders to automatically use the mock mode?

PR #200, opened 2020-08-24
KMS support for Always free and Advanced Options to allow change Shape and support MSZ

junior:
On this PR:
- Simplified view for first time users
- Advanced Options option
- Change the standard SSH keys creation moved to Advanced Options
- Option to Encrypt in Transit for the Compute <> Boot Volume
- OCI Vault (KMS) support to encrypt boot volume and object storage. NOTE: KMS is available on the Always free
- Option to create OCI Vault and Key automatically or use existent KMS Key
- Policy creation automated
- Option to create KMS policy for the user group (use case: someone have access to manage all resources in a compartment, but do not have specific policies to manage vault and keys)
- Option to not force use only Always Free eligible resources
- Option to change the shape of Load Balancer
- Option to change the compartment for the Load Balancer
- Option to change the shape, image operating system and version for Compute
- Option to use Always-free ATP or not
- Option to change the compartment of the Object Storage for the MuShop media
- Initial support for the OCI Maximum Security Zones, Maximum Security Recipe - 2020Q4

image

Test release: 
[mushop-basic-stack-v1.2.0.beta.1.zip](https://github.com/oracle-quickstart/oci-cloudnative/files/5119498/mushop-basic-stack-v1.2.0.beta.1.zip)


deploy/basic/terraform/outputs.tf
```diff
@@ -19,11 +19,17 @@ output "generated_private_key_pem" {
 }
 
 output "dev" {
-  value = "Made with \u2764 by Oracle A-Team"
+  value = "Made with \u2764 by Oracle A-Team Alumni"
```
mvandervliet, line 22: 😆 .  But...

junior: accepting suggestions that are more close to reality.

deploy/basic/terraform/scripts/node.sh
```diff
@@ -38,7 +38,7 @@ yum install -y gcc-c++ make nodejs wget unzip httpd jq
 
 yum -y install oracle-release-el7
 yum-config-manager --enable ol7_oracle_instantclient
-yum -y install oracle-instantclient19.3-basic oracle-instantclient19.3-jdbc oracle-instantclient19.3-sqlplus
+yum -y install oracle-instantclient8-basic oracle-instantclient19.8-jdbc oracle-instantclient19.8-sqlplus
```
mvandervliet, line 41: is this right? `instantclient8`?

src/storefront/src/templates/partials/_footer.pug
```diff
@@ -62,7 +62,7 @@ section.uk-section.uk-section-secondary.uk-section-xsmall.uk-light
         div.uk-margin-small.uk-text-muted.uk-text-small
           |Made with
           +icon('heart', '0.75')(class="uk-margin-xsmall-left uk-margin-xsmall-right")
-          |by Oracle A-Team
+          |by Oracle A-Team Alumni
```
mvandervliet, line 65: I like it... laughed... but don't think we should do it

deploy/basic/terraform/scripts/node.sh
junior: Noticed that after the PR. Was not affecting the usage as the jdbc force to instal the correct basic. Fixed anyways.

src/storefront/src/templates/partials/_footer.pug
junior: But... but... the badges 🤣

deploy/basic/terraform/provider.tf
```diff
@@ -1,11 +1,16 @@
...
+  # required_providers {
+  #   oci = ">= 3.90"
+  #   tls = ">= 2.0"
+  #   random = ">= 2.1"
+  # }
+}
+
 provider "oci" {
   tenancy_ocid = var.tenancy_ocid
-  # user_ocid        = "${var.user_ocid}"
```
jeevanjoseph, line 7: I know these are commented out, but the comments help people to run this locally without Resource manager. 
I think there is value in that.

src/storefront/src/templates/partials/_footer.pug
jeevanjoseph: :+1

deploy/basic/terraform/datasources.tf
```diff
@@ -58,4 +71,16 @@ data "oci_core_images" "compute_images" {
...
   sort_order               = "DESC"
-}
\ No newline at end of file
+}
+
+data "template_file" "mushop" {
+  template = "${file("./scripts/node.sh")}"
+}
+
+locals {
```
jeevanjoseph, line 80: nit: Move `locals` to a single location or have them dispersed to the locations where they are used. 
This one is used in `compute.tf`, but defined here. There are locals in `policies.tf`, but used within `policies.tf`

deploy/basic/terraform/policies.tf
```diff
@@ -0,0 +1,43 @@
...
+  compartment_id = var.compartment_ocid
+  statements = (var.use_encryption_from_oci_vault ?
+    (var.create_vault_policies_for_group ? (
+      concat(local.allow_object_storage_lifecycle_statement, local.allow_object_storage_service_keys_statements, local.allow_group_manage_vault_keys_statements)
+      ) : concat(local.allow_object_storage_lifecycle_statement, local.allow_object_storage_service_keys_statements)
+  ) : local.allow_object_storage_lifecycle_statement)
+  freeform_tags = local.common_tags
+}
+
+locals {
```
jeevanjoseph, line 33: Same comment about `locals`

deploy/basic/Dockerfile
```diff
@@ -150,18 +148,19 @@ RUN chmod +x /entrypoint.sh
...
 COPY --from=storefront-builder /app/storefront/build /app/storefront
 COPY --from=api-builder /app/api /app/api
 COPY --from=catalogue-builder /catalogue /app/catalogue/catalogue
+COPY src/catalogue/VERSION /app/catalogue/
 
 # Create zip package of the Apps and local images
 RUN mkdir /package && GZIP=-9 tar cvzf /package/mushop-basic.tar.gz /app
 
 # Create ORM package
+COPY deploy/basic/VERSION /
```
jeevanjoseph, line 163: 👍

deploy/basic/terraform/outputs.tf
junior: After voting, we will keep like this for now. Can re-visit later.

src/storefront/src/templates/partials/_footer.pug
junior: After voting, we will keep like this for now. Can re-visit later.

deploy/basic/terraform/policies.tf
junior: Here are the "policy" locals that only serve the policies.tf

deploy/basic/terraform/datasources.tf
junior: The locals on "datasources" are in a central location, with locals used by multiple resources. The locals in policies.tf are just used by policies.tf.

deploy/basic/terraform/provider.tf
junior: Yes. forgot to change. Will change to support both, ORM and local. Can completely remove as was before, as the main usage of mushop basic is ORM stack

junior: I've included because the ORM schema was not validating because do not had that entries.

deploy/basic/terraform/policies.tf
junior: locals moved per recommendation

deploy/basic/terraform/datasources.tf
junior: locals moved per recommendation

deploy/basic/terraform/provider.tf
junior: Fixed. Variables included to avoid ORM internal validation errors. Used default="" to be workable on ORM or plain terraform.

deploy/basic/terraform/policies.tf
```diff
@@ -0,0 +1,43 @@
+# Copyright (c) 2019, 2020 Oracle and/or its affiliates. All rights reserved.
+# Licensed under the Universal Permissive License v 1.0 as shown at http://oss.oracle.com/licenses/upl.
+# 
+
+# Create lifecycle policy to delete temp files
+resource "oci_objectstorage_object_lifecycle_policy" "mushop_deploy_assets_lifecycle_policy" {
```
mvandervliet, line 6: IIRC this was causing problems for us initially. Has this been verified?

junior: yes. Is being correctly generated and on the right place. If someone has access to a specific compartment and not to the tenancy, now works fine.

PR #214, opened 2020-11-01
Observability on Oracle Cloud Infrastructure (OCI) by simulating a failure scenario on MuShop application.

naikvenu:
This Scenario covers various Oracle Cloud Infrastructure(OCI) Observability Services.

Details:
    User logs in to MuShop and Orders items with value more than $105.
    User notices the Order processing failure.
    User gets notified on the Payment failures. 
    User analyzes the failure using Oracle Cloud Infrastructure(OCI) Observability Services (Logging, notification, alarms, monitoring etc)
    User delivers the fix.

Technical details:
Added a Business metric to send out failures to OCI monitoring using java sdk.
Added docs to cover the scenario (Logging, notification, alarms etc.)


jeevanjoseph (changes requested):
IMHO, this PR needs a rethinking on the architecture and approach here. 
From a micro services stand point, the orders service should not be doing any of this, but rather should be a separate service altogether purely focused on observability. This has a few advantages

- Avoids service bloat, and keeps the micro service focused on its core task.
- Services are more functionally resilient even when the OCI APIs fail or are unavailable
- Functionally decouples monitoring from order processing, and technically decouples the monitoring stack from the order processing path.
- Keeps order service decoupled from OCI. 
- Makes monitoring entirely optional (it should be optional) and pluggable (we should be able to replace that service with some other monitoring stack)

I suggest the following change in design and implementation.

1. Orders flow remains unchanged - Orders will post to payments, which can approve or deny a payment, based on its own logic.

2. Payment application emits a message on to a NATS subject to signal processing status for orders. (The current Orders<->Payment interaction is an API call, which can be updated to be message driven as well, which would be an improvement to that interface, but that's orthogonal really)

3. A new service subscribes to the NATS subject and detects the failed or successful transaction, the monitoring flow is implemented by the service, which can call OCI APIs. This service is coupled to OCI.

In an alternate implementation(hypothetical), this micro service can be replaced with a similar service for say, cloud watch, to support AWS native APIs without impacting the functional micro services in the service graph.

src/orders/Dockerfile
```diff
@@ -3,20 +3,57 @@ ARG APPLICATION_NAME="orders"
...
+
+# COPY --from=oracleinstantclientjdbc /usr/lib/oracle/${JDBC_DRIVER_VERSION}/client64/lib/ojdbc8.jar /usr/lib/oracle/${JDBC_DRIVER_VERSION}/client64/lib/ojdbc8.jar
+# RUN mvn install:install-file -Dfile=/usr/lib/oracle/${JDBC_DRIVER_VERSION}/client64/lib/ojdbc8.jar -DgroupId=${JDBC_GROUP_ID} -DartifactId=${JDBC_DRIVER_ARTIFACT_ID} -Dversion=${JDBC_DRIVER_VERSION} -Dpackaging=jar
+
+# Setup OCI libraries
+ARG GROUP_ID="com.oracle.oci"
+ARG ARTIFACT_ID="oci-java-sdk"
+ARG ARTIFACT_VERSION="1.25.1"
+
+ADD https://github.com/oracle/oci-java-sdk/releases/download/v${ARTIFACT_VERSION}/${ARTIFACT_ID}-${ARTIFACT_VERSION}.zip /root/
```
jeevanjoseph, line 26: This massively complicated builds for all users, even those that are not interested in the monitoring stack,

src/orders/README.md
```diff
@@ -206,6 +206,39 @@ This creates two secrets managed by kubernetes, named `atp-db-creds` and `atp-jk
...
+
+The above secret is required to send out custom metric (payment failure) to OCI Monitoring using OCI Java SDK.
+
+### For Developer setup:
+
+# Build
+
+## Java: building application jar
+
+OCI-java-sdk is a dependency and needs to be manually uploaded to
```
jeevanjoseph, line 229: Like before - this adds significant friction to the build process and all users are now forced to do this. We should likely split off the monitoring in to a separate service that's optional

src/orders/build.gradle
```diff
@@ -45,6 +45,26 @@ dependencies {
   }
 
   testImplementation 'com.openpojo:openpojo:0.8.12'
+ 
+  // Newly added dependencies for OCI SDK
+  implementation 'org.springframework.boot:spring-boot-starter-actuator'
```
jeevanjoseph, line 50: Are all these deps really required ?

src/orders/src/main/java/mushop/orders/config/MonitoringConfiguration.java
```diff
@@ -0,0 +1,163 @@
...
+        final String userId = env.getProperty("OCI_USER_ID");
+        final String fingerprint = env.getProperty("OCI_FINGERPRINT");
+        final String privateKey = env.getProperty("OCI_API_KEY");
+        final String passPhrase = env.getProperty("OCI_PASSPHRASE");
+        final String monitoringEndPoint = env.getProperty("OCI_MONITORING_ENDPOINT");
+        compartmentId = env.getProperty("OCI_COMPARTMENT_ID");
+        namespace = "mushopnamespace";
+        region = env.getProperty("OCI_REGION");
+    
+        System.out.println("Started Initializing the Monitoring Service...");
```
jeevanjoseph, line 72: Use the logger.

src/orders/src/main/java/mushop/orders/config/MonitoringConfiguration.java
```diff
@@ -0,0 +1,163 @@
...
+        //System.out.printf("%s,%s,%s,%s,%s,%s",tenantId,userId,fingerprint,privateKey,monitoringEndPoint,compartmentId);
+        
+        if (tenantId == null || userId == null || fingerprint == null || privateKey == null){
+                System.out.printf("Cannot send metrics to OCI Monitoring, Ensure you have set the required ENV variables tenantId, userId, fingerprint and privateKey ..");
```
jeevanjoseph, line 76: logger

src/orders/src/main/java/mushop/orders/config/MonitoringConfiguration.java
```diff
@@ -0,0 +1,163 @@
...
+        return monitoringClient;
+    }
+
+    public static void post(
+            MonitoringClient monitoringClient,
+            String compartment,
+            String namespace,
+            String metricName,
+            String region) {
+        final PostMetricDataRequest request =
```
jeevanjoseph, line 117: nit: formatting. avoid tabs, 2 space indentations.

src/orders/src/main/java/mushop/orders/config/MonitoringConfiguration.java
```diff
@@ -0,0 +1,163 @@
...
+                                                                        makeMap(
+                                                                                "region",
+                                                                                    region,
+                                                                                "pod-name",
+                                                                                  "mushop-orders"))
+                                                                .build()))
+                                        .build())
+                        .build();
+
+        System.out.printf("Request constructed:\n%s\n\n", request.getPostMetricDataDetails());
```
jeevanjoseph, line 146: logger across the board.

src/orders/src/main/java/mushop/orders/config/MonitoringConfiguration.java
```diff
@@ -0,0 +1,163 @@
...
+                .region(Region.fromRegionId(region))
+                .build();
+        }
+                
+        monitoringClient = new MonitoringClient(provider);
+
+        if (monitoringEndPoint != null){
+                monitoringClient.setEndpoint(monitoringEndPoint);
+        } else {
+                monitoringClient.setEndpoint("https://telemetry-ingestion.us-phoenix-1.oraclecloud.com");
```
jeevanjoseph, line 106: externalize configurations and endpoints

jeevanjoseph: Also - docs to update that when deployng to regions other than Phx.

src/orders/src/main/java/mushop/orders/config/MonitoringConfiguration.java
```diff
@@ -0,0 +1,163 @@
...
+                provider = SimpleAuthenticationDetailsProvider.builder()
+                .tenantId(tenantId)
+                .userId(userId)
+                .fingerprint(fingerprint)
+                .privateKeySupplier(new StringPrivateKeySupplier(privateKey))
+                .region(Region.fromRegionId(region))
+                .build();
+        }
+                
+        monitoringClient = new MonitoringClient(provider);
```
jeevanjoseph, line 101: We should likely split this off in to a separate service - Orders having a dependency on this client seems too bloated.

src/orders/src/main/java/mushop/orders/services/OrdersService.java
```diff
@@ -24,6 +24,9 @@
 import static mushop.orders.controllers.OrdersController.OrderFailedException;
 import static mushop.orders.controllers.OrdersController.PaymentDeclinedException;
 
+import mushop.orders.config.MonitoringConfiguration;
+import com.oracle.bmc.monitoring.MonitoringClient;
```
jeevanjoseph, line 28: Orders Service depending on MonitoringClient seems too tightly coupled to OCI and a very specific monitoring stack. Should decouple the implementation, or more likely spin off in to a separate async service

src/orders/src/main/java/mushop/orders/services/OrdersService.java
```diff
@@ -93,6 +100,23 @@ public CustomerOrder createNewOrder(NewOrderResource orderPayload) {
             }
             if (!paymentResponse.isAuthorised()) {
                 meterRegistry.counter("orders.rejected","cause","payment_declined").increment();
+                // Send Payment failure metrics to OCI Monitoring..
+                try {
+                    System.out.println("In Order Service, Setting up OCI Monitoring .. ");
```
jeevanjoseph, line 105: More System.outs ... use the logger

src/orders/src/main/java/mushop/orders/services/OrdersService.java
```diff
@@ -93,6 +100,23 @@ public CustomerOrder createNewOrder(NewOrderResource orderPayload) {
...
+                    monitoringClient = monitoringConfig.initConnection();
+                    if (monitoringClient == null) {
+                        System.out.printf("Cannot send metrics to OCI Monitoring, Ensure you have set the required ENV variables tenantId, userId, privateKey and fingerprint ..");
+                    }else {
+                        String compartment = monitoringConfig.getCompartmentId();
+                        String namespace = monitoringConfig.getNamespace();
+                        String MetricName = "Payment-Failure";
```
jeevanjoseph, line 112: externalize final strings as constants.

src/orders/src/main/java/mushop/orders/services/OrdersService.java
```diff
@@ -93,6 +100,23 @@ public CustomerOrder createNewOrder(NewOrderResource orderPayload) {
             }
             if (!paymentResponse.isAuthorised()) {
                 meterRegistry.counter("orders.rejected","cause","payment_declined").increment();
+                // Send Payment failure metrics to OCI Monitoring..
+                try {
+                    System.out.println("In Order Service, Setting up OCI Monitoring .. ");
+                    monitoringClient = monitoringConfig.initConnection();
+                    if (monitoringClient == null) {
+                        System.out.printf("Cannot send metrics to OCI Monitoring, Ensure you have set the required ENV variables tenantId, userId, privateKey and fingerprint ..");
```
jeevanjoseph, line 108: monitoring client could be wrapped in an Option(MonitoringClient) and case matched ?
But this service is doing way more than order processing at this point.

src/docs/content/observability/observability-scenario/pre-requisites/oci-notifications.md
```diff
@@ -0,0 +1,31 @@
+---
```
jeevanjoseph, line 1: This pre-req should be optional. In many settings users may not have the capacity and makes the "getting started" curve steep.

src/docs/content/observability/observability-scenario/pre-requisites/oci-logging.md
```diff
@@ -0,0 +1,106 @@
+---
```
jeevanjoseph, line 1: This pre-req should be optional. In many settings users may not have the capacity and makes the "getting started" curve steep.

src/docs/content/observability/observability-scenario/_index.md
```diff
@@ -0,0 +1,22 @@
...
+In this section we will try to understand Observability on Oracle Cloud Infrastructure (OCI) by simulating a failure scenario on MuShop application.
+
+![Failure-scenario](../images/observability-scenario.png)
+
+## Scenario Details
+- User logs in to MuShop and Orders items with value more than $105.
+- User notices the Order processing failure.
+- User gets notified on the Payment failures. 
+  In a production setup, application operations team would get paged.
+- User analyzes the failure using Oracle Cloud Infrastructure(OCI) Observability Services.
```
jeevanjoseph, line 19: Using the observability service to monitor for a functional scenario like this seems a bit unnatural. For instance, perhaps the Monitoring tool can detect say N failed attempts within P seconds and trigger a fraud alert that pages ops to disable the user ?

jeevanjoseph (changes requested):
Some comments - but overall I think this is better implemented in a decoupled and optional manner, so that those users who want to use this track can do so. The service bloat should also be avoided by sticking to micro service practices of keeping the service singularly focused, and light footed. see comment above.

deploy/complete/kubernetes/README.md
```diff
@@ -11,8 +11,26 @@ Refer to [helm-chart](../helm-chart/README.md)
 > NOTE: There are several options for ingress controllers in K8S. This demo uses the common `ingress-nginx`
 
 ```text
-kubectl apply -f https://raw.githubusercontent.com/kubernetes/ingress-nginx/master/deploy/static/mandatory.yaml
-kubectl apply -f https://raw.githubusercontent.com/kubernetes/ingress-nginx/master/deploy/static/provider/cloud-generic.yaml
+kubectl apply -f https://raw.githubusercontent.com/kubernetes/ingress-nginx/nginx-0.30.0/deploy/static/mandatory.yaml
```
junior, line 14: We should not alter this and hardcode a specific version. We already have an effort to always support the latest.

deploy/complete/kubernetes/README.md
```diff
@@ -11,8 +11,26 @@ Refer to [helm-chart](../helm-chart/README.md)
...
+```
+
+```bash
+kubectl -n ingress-nginx get svc
+```
+
+### Enable the monitoring capability by adding your OCI credentials:
+
+```bash
+kubectl create secret generic monitoring-secret \
```
junior, line 25: Create a OCI secret need to be optional, as we support deploy MuShop on any platform and in mock mode, not needing any extra configuration.

deploy/complete/kubernetes/mushop.yaml
```diff
@@ -1216,7 +1216,8 @@ spec:
         
       containers:
         - name: orders
-          image: "iad.ocir.io/oracle/ateam/mushop-orders:2.2.2"
+          #image: "iad.ocir.io/oracle/ateam/mushop-orders:2.2.2"
+          image: "phx.ocir.io/ociateam/mushop-orders:4.0"
```
junior, line 1220: We cannot use non-official tenancies, and gets worse if you change the main manifest with that. Our CI automatically pushes the Microservices image to the official OCIR, if you think that is not using the latest, change just the version.

deploy/complete/kubernetes/mushop.yaml
```diff
@@ -1216,7 +1216,8 @@ spec:
         
       containers:
         - name: orders
-          image: "iad.ocir.io/oracle/ateam/mushop-orders:2.2.2"
+          #image: "iad.ocir.io/oracle/ateam/mushop-orders:2.2.2"
```
junior, line 1219: There's no reason to include commented entries on the manifest. If is not using it, delete it.

deploy/complete/kubernetes/mushop.yaml
```diff
@@ -1239,7 +1240,7 @@ spec:
             - name: ZIPKIN
               value: zipkin.jaeger.svc.cluster.local
             - name: JAVA_OPTS
-              value: -Xms32m -Xmx150m -XX:MaxRAM=150m -Djava.security.egd=file:/dev/urandom -Doracle.jdbc.fanEnabled=false -XX:+UnlockExperimentalVMOptions -XX:+UseZGC -Dlogging.level.mushop.orders=INFO -Dspring.zipkin.enabled=false
+              value: -Xms64m -Xmx300m -XX:MaxRAM=300m -Djava.security.egd=file:/dev/urandom -Doracle.jdbc.fanEnabled=false -XX:+UnlockExperimentalVMOptions -XX:+UseZGC -Dlogging.level.mushop.orders=INFO -Dspring.zipkin.enabled=false
```
junior, line 1243: The memory usage is already high, and is needed to low to fit lower shapes and preparation for Always-free support. If need to increase for a specific use case, need to include as variable and change on demand, keeping the low entry as default.

deploy/complete/kubernetes/mushop.yaml
```diff
@@ -1263,6 +1264,48 @@ spec:
                 secretKeyRef:
                   name: orders-oadb-connection
                   key: oadb_service
+                  
+            - name: OCI_TENANT_ID
```
junior, line 1268: External entries like OCI credentials and specific entries for a tenancy should never be mandatory.

src/docs/content/observability/observability-scenario/pre-requisites/setup.md
```diff
@@ -0,0 +1,63 @@
...
+## Pre-requisites
+
+### Install Ingress Controller
+
+```bash
+kubectl create namespace mushop
+```
+```text
+
+kubectl apply -f https://raw.githubusercontent.com/kubernetes/ingress-nginx/nginx-0.30.0/deploy/static/mandatory.yaml
```
junior, line 19: We should not include specific version. This including is an old version that is not compatible with all features from other services.

src/docs/content/observability/observability-scenario/pre-requisites/setup.md
```diff
@@ -0,0 +1,63 @@
+---
+title: "Setup"
+date: 2020-03-10T13:29:23-06:00
+draft: false
+weight: 1
+tags:
+  - mushop setup
+---
+
+## Pre-requisites
```
junior, line 10: This prerequisites on the observability level, makes the visitors think that they need to do all these steps to be able to have basic observability, this is not true. These are optional steps for a specific use-case.

src/docs/content/observability/observability-scenario/pre-requisites/setup.md
```diff
@@ -0,0 +1,63 @@
...
+
+### Install Ingress Controller
```
junior, line 12: Create an ingress manually to a standard deployment (Which already includes an Ingress Nginx) will cause conflicts and may break the solution. This also is not the same package, causing more disturbance on the deployment.

src/docs/content/observability/observability-scenario/pre-requisites/setup.md
```diff
@@ -0,0 +1,63 @@
...
+```text
+
+kubectl apply -f https://raw.githubusercontent.com/kubernetes/ingress-nginx/nginx-0.30.0/deploy/static/mandatory.yaml
+kubectl apply -f https://raw.githubusercontent.com/kubernetes/ingress-nginx/nginx-0.30.0/deploy/static/provider/cloud-generic.yaml
+```
+
+### Create a Secret for OCI Monitoring
+
+```bash
+kubectl create secret generic monitoring-secret \
```
junior, line 26: Include OCI credentials should never be something mandatory. Also, there's a lot of intelligence on the scripts to reuse ocicredentials secrets. Create a separate secret for a specific service should be a second optional option.

src/docs/content/observability/observability-scenario/pre-requisites/setup.md
```diff
@@ -0,0 +1,63 @@
...
+```
+
+## Deploy
+
+```
+cd deploy/complete/kubernetes
+```
+
+```text
+kubectl apply -f mushop.yaml
```
junior, line 44: Why suggesting to use regular manifest and not helm charts?

src/orders/Dockerfile
```diff
@@ -3,20 +3,57 @@ ARG APPLICATION_NAME="orders"
 ARG VERSION="0.0.1-SNAPSHOT"
 
 # ------------
-# Stage 1 : Setup the build environment
-FROM gradle:6.5 as buildenv
+# Stage 1 : Setting up the build environment with maven and gradle
+# 
+# 
+FROM maven:3.6-jdk-8 as buildenv
```
junior, line 9: Why this old way to build come back?

src/orders/Dockerfile
```diff
@@ -3,20 +3,57 @@ ARG APPLICATION_NAME="orders"
...
+
+ARG JDBC_DRIVER_VERSION
+ARG JDBC_GROUP_ID="com.oracle.jdbc"
+ARG JDBC_DRIVER_ARTIFACT_ID="ojdbc8"
+ARG GRADLE_VERSION=5.6.1
+
+ENV GRADLE_HOME /opt/gradle
+
+# COPY --from=oracleinstantclientjdbc /usr/lib/oracle/${JDBC_DRIVER_VERSION}/client64/lib/ojdbc8.jar /usr/lib/oracle/${JDBC_DRIVER_VERSION}/client64/lib/ojdbc8.jar
```
junior, line 18: What is the reason to have commented code on the scripts?

src/orders/README.md
```diff
@@ -206,6 +206,39 @@ This creates two secrets managed by kubernetes, named `atp-db-creds` and `atp-jk
 The kubernetes deployment manifest `atp-spring-jpa.yml` references these secrets and provides them to the runtime environment of the containers. The application now references the properties through the environment variables instead of reading them from a file. Alternatively the config files themselves could have been provided and mounted as volumes in to the container.
 
 
+```bash
+kubectl create secret generic monitoring-secret \
```
junior, line 210: WTH? Why makes mandatory include this secret? This basically breaks several use case usages of MuShop, breaking several labs, workshops that are already in use. This need to be something totally optional

src/orders/README.md
```diff
@@ -206,6 +206,39 @@ This creates two secrets managed by kubernetes, named `atp-db-creds` and `atp-jk
...
+### For Developer setup:
+
+# Build
+
+## Java: building application jar
+
+OCI-java-sdk is a dependency and needs to be manually uploaded to 
+the local maven repository. 
+* download oci-java-sdk 
+`wget https://github.com/oracle/oci-java-sdk/releases/download/v1.25.1/oci-java-sdk-1.25.1.zip`
```
junior, line 232: image

We spend a lot of time automating everything to not create any manual step or manual dependency. This should be corrected and if really needed, need to be automated. NEVER be a "local maven repository" or a manual step. This also breaks our current CI.

src/orders/build.gradle
junior: +1 (Why?)

src/orders/src/main/java/mushop/orders/config/MonitoringConfiguration.java
```diff
@@ -0,0 +1,163 @@
...
+
+    public String getNamespace() {
+	return namespace;
+    }
+
+    public String getCompartmentId() {
+	return compartmentId;
+    }
+    
+    public MonitoringClient initConnection() throws Exception {
```
junior, line 60: This breaks the current usage. Orders should be able to run independently from any manual requirements and of course, the worse, need to create credentials to be used.

src/orders/src/main/java/mushop/orders/services/OrdersService.java
```diff
@@ -24,6 +24,9 @@
 import static mushop.orders.controllers.OrdersController.OrderFailedException;
 import static mushop.orders.controllers.OrdersController.PaymentDeclinedException;
 
+import mushop.orders.config.MonitoringConfiguration;
```
junior, line 27: image

junior (changes requested):
Why are you changing things on /kubernetes ??? These files are automatically generated from `helm`, would be replaced or out of sync on the next update. So, should not be used for any update.

jeevanjoseph (changes requested):

src/orders/src/main/java/mushop/orders/config/MonitoringConfiguration.java
jeevanjoseph: +1

deploy/complete/kubernetes/mushop.yaml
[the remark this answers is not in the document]
naikvenu: I faced java heap space issues with 150M and had to increase it.

src/orders/src/main/java/mushop/orders/services/OrdersService.java
naikvenu: ha ha :-)

src/orders/src/main/java/mushop/orders/config/MonitoringConfiguration.java
naikvenu: Agreed. BTW, credentials are totally optional, the doc needs to be corrected. If no credentials set, the code just ignores sending the metrics out to OCI monitoring.

naikvenu: implementation 'org.springframework.boot:spring-boot-starter-actuator'

All those dependencies were taken from shipping service. Its was a bit of trial and error. Let me check on this.

src/orders/README.md
naikvenu: This is just for developer setup running on your laptop. Everything is automated using Dockerfile for production.

src/orders/README.md
naikvenu: This is not a mandatory step.. purely optional. I will correct the doc.

src/orders/Dockerfile
naikvenu: I can remove that.

src/orders/Dockerfile
naikvenu: To accomodate SDK.

src/docs/content/observability/observability-scenario/pre-requisites/setup.md
naikvenu: To use enhanced mock mode which captures communication between different services and logging excluding catalogue. we needed this to implement the payment failure scenario.
the basic mock is only limited to API service.

src/docs/content/observability/observability-scenario/pre-requisites/setup.md
naikvenu: It is an optional step. I will correct doc.

src/docs/content/observability/observability-scenario/pre-requisites/setup.md
naikvenu: we need to deploy ingress controller before applying mushop.yaml. which then applies the rules from ingress to edge.

src/docs/content/observability/observability-scenario/pre-requisites/setup.md
naikvenu: this pre-requisite is under the observability-scenario

src/docs/content/observability/observability-scenario/pre-requisites/setup.md
naikvenu: took it from here: https://docs.cloud.oracle.com/en-us/iaas/Content/ContEng/Tasks/contengsettingupingresscontroller.htm

deploy/complete/kubernetes/mushop.yaml
naikvenu: Hmm, agree. I would suggest mushop-monit-scenario.yaml

deploy/complete/kubernetes/mushop.yaml
naikvenu: I had faced java heal space issue and had to increase. I am considering to implement new service.

deploy/complete/kubernetes/mushop.yaml
naikvenu: was waiting for the build version to add.

deploy/complete/kubernetes/mushop.yaml
naikvenu: yes agree, this is just a placeholder for now.

deploy/complete/kubernetes/README.md
naikvenu: It is optional, however will need to change mushop.yaml to mushop-monit-scenario.yaml and a new README if we decide to go ahead with this implementation. Can remove these steps from README and just add it on the scenario.md would also help avoid confusions.

mvandervliet:
As a general consideration: what would you think about using the events service for this instead? It already has optional (with noop) credentials and OCI sdk in go. Could easily tie into the monitoring as well

deploy/complete/kubernetes/README.md
naikvenu: kubectl apply -f https://raw.githubusercontent.com/kubernetes/ingress-nginx/master/deploy/static/mandatory.yaml
doesnt work anymore. so had to change.

took it from here: https://docs.cloud.oracle.com/en-us/iaas/Content/ContEng/Tasks/contengsettingupingresscontroller.htm
we can also use kube-proxy to forward to edge. but optionally can use ingress also. that was my intent.

naikvenu: is this generated by helm ?

src/docs/content/observability/observability-scenario/_index.md
naikvenu: The idea for initial release was to just show that domain specific business metrics are possible using a simple payment failure situation and alerting plus skimming through the logs to find out what happened. We can keep building on top of it as we go. fyi: with monitoring we can only send out time based data and we do not send any user id's there.

src/docs/content/observability/observability-scenario/pre-requisites/oci-logging.md
naikvenu: idea was to have users understand active and passive monitoring (looking at OCI monitoring dashboard and getting alerts via alarms.) Plus looking at logs to root cause the issue.
How can OCI observability help determine failure to observe to action. This was the intent. I agree its a bit of a curve.

src/docs/content/observability/observability-scenario/pre-requisites/oci-notifications.md
naikvenu: idea was to have users understand active and passive monitoring (looking at OCI monitoring dashboard and getting alerts via alarms.) Plus looking at logs to root cause the issue.
How can OCI observability help determine failure -> observe -> action. This was the intent. I agree its a bit of a curve.

src/orders/src/main/java/mushop/orders/services/OrdersService.java
naikvenu: I'll discuss on the new service option suggested.

src/orders/src/main/java/mushop/orders/services/OrdersService.java
naikvenu: ok

src/orders/src/main/java/mushop/orders/services/OrdersService.java
naikvenu: I think you meant to say monitoringClient here.

src/orders/src/main/java/mushop/orders/config/MonitoringConfiguration.java
naikvenu: ok

src/orders/src/main/java/mushop/orders/config/MonitoringConfiguration.java
naikvenu: 👍

src/orders/src/main/java/mushop/orders/config/MonitoringConfiguration.java
naikvenu: 👍

src/orders/src/main/java/mushop/orders/config/MonitoringConfiguration.java
naikvenu: got it

src/orders/src/main/java/mushop/orders/config/MonitoringConfiguration.java
naikvenu: actually the monitoring-secret collects monitoring_endpoint
 
kubectl create secret generic monitoring-secret \
--from-literal=compartment_id="<your compartment_id>" \
--from-literal=tenant_id="<your tenancy_id>" \
--from-literal=user_id="<your user_id>" \
--from-literal=fingerprint="<your fingerprint>" \
--from-literal=passphrase="<your passphrase>" \
--from-literal=monitoring_endpoint="<your monitoring endpoint>" \
--from-literal=region="<your region>" \
--from-file=apikey=<your api_key.pem>

src/orders/build.gradle
naikvenu: they are taken from shipping service as is.. bit of trial and error here. I'll revisit this.

src/orders/README.md
naikvenu: that sounds like a better option.

src/orders/Dockerfile
naikvenu: 👍

src/orders/README.md
naikvenu: this is only for dev env on laptop. For user's it is pre-built (DockerFile)

src/orders/src/main/java/mushop/orders/services/OrdersService.java
naikvenu: yes. I was thinking if a well tested image would be just good ex: iad.ocir.io/oracle/ateam/mushop-orders-monit:2.2.3

deploy/complete/kubernetes/README.md
junior: yes. The main deployment are on deploy/complete/helm-charts. deploy/complete/kubernetes are generated from there.

The setup chart that takes care of the ingress, and has specific configurations to support Grafana and other services. We do not use the suggested standard from the documentation.

junior: And we nginx 0.30 is deprecated, and has some documented issues. Even for other examples, you should be using 3.x+ instead of 0.x

deploy/complete/kubernetes/README.md
junior: "monit" sounds strange, and do not need to specify that is a "Scenario". Use mushop_with_monitoring.yaml or something more verbal.

deploy/complete/kubernetes/mushop.yaml
junior: a placeholder on the main mushop values (deploy/complete/kubernetes/mushop.yaml). Need to change before merge.

deploy/complete/kubernetes/mushop.yaml
junior: Change the PR to WIP, so we wait to comment.

deploy/complete/kubernetes/mushop.yaml
junior: 150m is actually too high. Orders is the bigger offender, catalogue for example uses 17Mb in load. MuShop runs on micro shapes and externally on small infra (1.1, Raspberry PI, arm, ...)

src/orders/README.md
junior: The developer machine, CI and server should not be different. The build process should be the same, without manual steps.

deploy/complete/kubernetes/mushop.yaml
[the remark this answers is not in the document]
junior: Orders should be under 128m by now. The only excuse is that is 2020 and COVID-19.

naikvenu:
> As a general consideration: what would you think about using the events service for this instead? It already has optional (with noop) credentials and OCI sdk in go. Could easily tie into the monitoring as well

it makes sense to me if variables related to stream are not mandatory. I see STREAM_ID is required. I too agree New service seems more appropriate as the jar is a bit big.
I started with node.js as mock is implemented in that, later when we decided to use enhanced mock, the focus turned to order service. Due to this, I couldnt forsee how fundamentally it could violate micro services principles.

deploy/complete/kubernetes/mushop.yaml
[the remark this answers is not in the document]
naikvenu: 👍

naikvenu:
Closing this PR, and will open a new PR with better design.

PR #216, opened 2020-11-09
Allow docs tests on forked repo

junior:
- Allow relative URLs, enabling publish on gh-pages on forks
- remediate empty baseURL links
- enable workflow on repo named `docs_update` (protected on main repo) to publish to gh-pages on forked

Example from this PR: https://junior.github.io/oci-quickstart-cloudnative/

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches: 
     - master
+    - docs_update
     paths:
     - 'src/docs/**'
 jobs:
--- a/src/docs/README.md
+++ b/src/docs/README.md
@@ -42,4 +42,4 @@ hugo server --buildDrafts
 
 ### Usage
 
-Open [http://localhost:3131/oci-coudnative/usage](http://localhost:3131/oci-coudnative/usage)
+Open [http://localhost:1313/usage](http://localhost:1313/usage)
--- a/src/docs/config.toml
+++ b/src/docs/config.toml
@@ -3,10 +3,12 @@
 # The Universal Permissive License (UPL), Version 1.0
 #
 baseURL = "https://oracle-quickstart.github.io/oci-cloudnative/"
+baseURL = ""
 languageCode = "en-us"
 title = "MuShop"
 theme = "redwood"
 relativeURLs = true
+# canonifyURLs = true
 
 [params]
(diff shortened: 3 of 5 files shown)


.github/workflows/docs.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches: 
     - master
+    - docs_update
```
mvandervliet, line 7: hmm... don't _love_ the idea of reusing the branch name over and over. but serves the purpose for your example

src/docs/config.toml
```diff
@@ -2,10 +2,11 @@
 # Copyright (c) 2020, Oracle and/or its affiliates. All rights reserved.
 # The Universal Permissive License (UPL), Version 1.0
 #
-baseURL = "https://oracle-quickstart.github.io/oci-cloudnative/"
+baseURL = ""
```
mvandervliet, line 5: can be omitted?

src/docs/README.md
```diff
@@ -42,4 +42,4 @@ hugo server --buildDrafts
 
 ### Usage
 
-Open [http://localhost:3131/oci-coudnative/usage](http://localhost:3131/oci-coudnative/usage)
+Open [http://localhost:1313/usage](http://localhost:1313/usage)
```
mvandervliet, line 45: doh

mvandervliet (approved):

src/docs/config.toml
junior: yes, maybe, trying later. :)

Maybe need to remove the Site.BaseURL requests from other places

.github/workflows/docs.yml
junior: I know. But is the best way to protect on main repo and execute the workflow. Open to suggestions.

src/docs/README.md
junior: No questions asked

src/docs/config.toml
mvandervliet: There is another approach that I've used for this - which is to configure the action with the `--baseURL` flag for `hugo` command.

junior: @mvandervliet I thought about that, but why? any benefits? I could not find any.

mvandervliet: nah not really. Thought it might be useful for the forks to match on other branches... then just incorporate logic into the action on whether or not to _actually_ deploy

PR #221, opened 2020-11-19
removed region defaults dependency

junior:
replaced with one console url for magic buttons and docs references.


mvandervliet (approved):
This whole time!

junior:
> This whole time!

Yeap. Finally. People around the world will not be mad that we "forced" then use Ashburn anymore. :)

PR #225, opened 2020-12-18
Catalogue service update

junior:
- Go updated to 1.15
- go libraries updated
- version bump
- using oraclelinux:8-slim
- initial support for multi arch build
- standalone manifest updated to the latest
- dbtools updated to the latest and versioning changed to match major Instant client


src/catalogue/Dockerfile
```diff
@@ -19,29 +21,35 @@ COPY *.go /go/src/mushop/catalogue/
...
-FROM oraclelinux:7-slim AS base
+FROM oraclelinux:8-slim AS base
 ARG clientVersion
-RUN yum update -y && \
-  yum -y install oracle-release-el7 && \
-  yum-config-manager --enable ol7_oracle_instantclient && \
-  yum -y install oracle-instantclient${clientVersion}-basiclite && \
-  yum clean all && \
-  rm -rf /var/cache/yum
+RUN microdnf update && \
```
mvandervliet, line 31: 😲

src/catalogue/Dockerfile
```diff
@@ -19,29 +21,35 @@ COPY *.go /go/src/mushop/catalogue/
...
+    microdnf clean all && \
+    rm -rf /var/cache/dnf && \
+    rm -rf /var/cache/yum
 
 ##### Catalogue Service Image
 FROM base
 ARG clientVersion
 WORKDIR /
-COPY --from=go-builder /app /app
+COPY --from=go-builder /catalogue /catalogue
```
mvandervliet, line 42: reason for the switch ~`/app`~ to `/catalogue`?

src/catalogue/kubernetes/catalogue-dep.yaml
```diff
@@ -32,7 +32,7 @@ spec:
         - name: wallet
           mountPath: /wallet
       - name: init-service-database
-        image: iad.ocir.io/oracle/ateam/mushop-dbtools:1.0.1
+        image: iad.ocir.io/oracle/ateam/mushop-dbtools:21
```
mvandervliet, line 35: So... I see where the `21` is used to represent the client version, and the `dbtools` is updated to reflect that. But, is that necessarily what we want to do? Might there be a case where "MuShop" would want to rev `dbtools` to 21.0.1 to update logic|scripts|whatever thus making it out of sync with the lexical client version? My feeling is that the dbtools version should just be a major bump to `2.0.0` in this case.

src/catalogue/Dockerfile
junior: No special reason, just matching the build of the basic.https://github.com/oracle-quickstart/oci-cloudnative/blob/a927f28e244f324ee76b468b295438c9ebe30552/deploy/basic/Dockerfile#L115

Will also help when changing the dbtools strategy.

src/catalogue/kubernetes/catalogue-dep.yaml
junior: dbtools just have the unmodified OracleDB tools, there are no scripts. The Instant Client changed the strategy to work with major version support, that's why skip 20, and 21 will last multiple version of the client and db.

If the InstantClient receives a major change, such as new authentication method, will change the major (22+).

So, the main binaries will still following the InstantClient/DB server major version strategy. And all changes can be tracked with minor.patch.

src/dbtools/Dockerfile
```diff
@@ -4,23 +4,25 @@
 # 
 
 ##### Oracle Instant Client version
-ARG clientVersion=19.5
+ARG clientVersion=21
```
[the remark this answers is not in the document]
junior: DB team changed the strategy, just following vibe :)

src/catalogue/Dockerfile
```diff
@@ -19,29 +21,35 @@ COPY *.go /go/src/mushop/catalogue/
...
 # Build Catalogue service
-RUN GOARCH=amd64 GOOS=linux \
+RUN GOARCH=${ARCH} GOOS=${OS} \
   go build -a \
-  -o /app mushop/catalogue/cmd/cataloguesvc
+  -o /catalogue mushop/catalogue/cmd/cataloguesvc
 
 ##### Base image with Oracle Instant Client Basic Lite
-FROM oraclelinux:7-slim AS base
+FROM oraclelinux:8-slim AS base
```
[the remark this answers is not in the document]
junior: image

src/catalogue/Dockerfile
mvandervliet: I see. IIRC the reason of `basic` doing this is to differentiate from the other services built within the same Dockerfile. 🤷‍♂️

src/catalogue/Dockerfile
```diff
@@ -4,10 +4,12 @@
 # 
 
 ##### Oracle Instant Client version
-ARG clientVersion=19.3
+ARG clientVersion=21
```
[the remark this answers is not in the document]
junior: image

src/catalogue/Dockerfile
junior: I can revert to the app be called app again for the app binary run as `app`. Didn't know that you loved so much.

src/catalogue/kubernetes/catalogue-dep.yaml
mvandervliet: > dbtools just have the unmodified OracleDB tools, there are no scripts. The Instant Client changed the strategy to work with major version support, that's why skip 20, and 21 will last multiple version of the client and db.

☝️ sort of. It still has this Dockerfile with installs and whatnot that we might need to change periodically. Theoretically the instant client can have its own semver that could differ. I wouldn't want there to be confusion as a result

src/catalogue/Dockerfile
mvandervliet: Don't ❤️ . Just wanted to understand the reason - even if it's just a "feels good" 😄 . I'm ok as-is

src/catalogue/kubernetes/catalogue-dep.yaml
junior: Oracle should provide that image, it's generic as is. Should only change if the InstantClient changes. There's no other reasons to make updates. We were using version 1.0.1 since the beginning and was using the old 19.3. The 1.0.2 was never used because was updated to 19.5 and break changes.

One of the reasons that the InstantClient changed the strategy, and started to use major version instead of major.minor, was exactly to avoid that issues, as some patches are necessary.

Including the version on the image tag, will also help when creating the volumes.

Some features are only available on the new DB client, and for the developer, `dbtools:2.0.0` will not tell anything, and the developer will be unaware why is not working. Will try to look at the dbtools:2.0.0 description on the docker repo, and will not see anything. Will just bring unnecessary time lost trying to debugging.

This is also some best practices that are being applied on other client tools.

src/catalogue/Dockerfile
junior: image

mvandervliet (approved):
Who's tasked with updating ALL the others?

junior:
> Who's tasked with updating ALL the others?

I'm looking to do that next week. That's why made the dbtools support both scenarios. Wanted to solve/fix this with one service before do all the rest.

PR #240, opened 2021-03-12
Updates, quick fixes and improvements

junior:
- Password DB profile for micro services without need to rotate passwords.
- Small fixes
- Small updates
- Helm Repo Repos updated
 - Terraform updated to 0.14
 - Terraform Providers updated
 - ORM Schema updated
 - Fixes and improvements
 - Supporting services updated
 - Support for more flex shapes


deploy/complete/helm-chart/mushop/charts/carts/config/atp.init.sql
```diff
@@ -1,9 +1,34 @@
...
+	BEGIN
+		SELECT COUNT(*) 
+		INTO profileExists 
+		FROM DBA_PROFILES 
+		WHERE PROFILE = 'MUSHOP_SERVICES';
+		DBMS_OUTPUT.PUT_LINE ('** Profile creationg steps - &_DATE');
+		IF profileExists = 0 THEN
+			DBMS_OUTPUT.PUT_LINE ('Creating Profile = MUSHOP_SERVICES ...');
+			EXECUTE IMMEDIATE
+				'CREATE PROFILE MUSHOP_SERVICES
```
mvandervliet, line 19: Do you think it would be appropriate here to match the profile according to the schema as a `1:1` rather than all services sharing the profile? I only say so since each of these init scripts will try to create the same profile. Feels a bit too replicated. Ideally it would be centralized instead, but I understand that's not too easy since each of these routines is performed at the service level.

deploy/complete/terraform/mushop.tf
```diff
@@ -19,7 +19,7 @@ resource "helm_release" "mushop" {
 
   set {
     name  = "global.mock.service"
-    value = var.mushop_mock_mode_all ? "all" : "false"
+    value = var.mushop_mock_mode_all ? "all" : "none"
```
mvandervliet, line 22: 👍

mvandervliet (approved):

deploy/complete/helm-chart/mushop/charts/carts/config/atp.init.sql
junior: I had the question. After some research, DBAs like to reduce number of profiles for easy management. Even on different DBs, they normally use the same name. When not using `DEFAULT`, the main profile name is normally `ALL_USERS`  and no much change after that. The granularity of the profiles goes just at Teams levels, when they have specific regulamentation. If is for an App or Service, is best practice to use a common name if there's no differences.

Said that, is also common practice to test for the requirements of the db procedure for each step, which is happening, and create the profile if does not exists.

Also, there's no way to guarantee who will execute first. Remember that we can also deploy one DB per service. Would need a lot of complexity and rules, just to avoid few lines on the SQL file.

deploy/complete/terraform/mushop.tf
junior: Yeah, Terraform 0.14 force the "false" to be boolean and test for the type.

PR #257, opened 2021-04-26
Basic - Adding changes for WAF, DNS, and automatic certificates

agregory999:
Please review - this functionality should be able to be added to the original "basic" configuration.


agregory999:
FYI, the NSG work is based on the work here -> https://github.com/BaptisS/oci_waf_nsg_v2

deploy/basic/terraform/variables.tf
```diff
@@ -192,3 +192,80 @@ locals {
...
+variable "waas_policy_waf_config_captchas_url" {
+  default = "/product.html"
+}
+
+variable "waas_policy_waf_config_captchas_footer_text" {
+  default = "Enter the letters and numbers as they are shown in the image above."
+}
+
+variable "waas_policy_waf_config_captchas_header_text" { 
+  default = "Help us keep pet product shopping safe please."
```
mvandervliet, line 269: Have you tested the MuShop SPA with CAPTCHA enabled? The SPA makes frequent `XHR` requests to the server for `*.html` files (views) during navigation. If one were to return the redirect for CAPTCHA, it might not be handled properly

deploy/basic/README_MSZ.md
```diff
@@ -21,10 +21,37 @@ If using local Terraform or on CloudShell:
 - Run `terraform init` to init the terraform providers
 - Run `terraform apply` to create the resources on OCI
 
-You can also include Web Application Firewall (WAF), DNS or other components as desired:
+You can also include Web Application Firewall (WAF), DNS or other components as desired.  Note that a public certificate will be required for WAF, and thus requires a DNS zone be created.  See below for more specific instructions.
 
 ![MuShop Basic Infra MSZ Demo](../../images/basic/00-Topology-v1.2.0.MSZ-demo.svg)
 
+### Enabling DNS, Certificates and WAF
```
junior, line 28: Should mention that these steps are optional

deploy/basic/terraform/cert.tf
```diff
@@ -0,0 +1,46 @@
+provider "acme" {
```
junior, line 1: Provider should be on providers.tf file.

Need to check if acme is supported by ORM

deploy/basic/terraform/cert.tf
```diff
@@ -0,0 +1,46 @@
...
+  server_url = var.acme_server_url
+}
+
+resource "tls_private_key" "private_key" {
```
junior, line 5: Already exist a resource with the same name

deploy/basic/terraform/cert.tf
```diff
@@ -0,0 +1,46 @@
...
+  algorithm = "RSA"
+}
+
+resource "acme_registration" "reg" {
```
junior, line 9: Should not have a dedicated file just for two resources

deploy/basic/terraform/cert.tf
```diff
@@ -0,0 +1,46 @@
...
+      OCI_TENANCY_OCID=var.tenancy_ocid
+      OCI_USER_OCID=var.user_ocid
+      OCI_PUBKEY_FINGERPRINT=var.fingerprint
+      OCI_REGION=var.region
+      OCI_COMPARTMENT_OCID=var.lb_compartment_ocid
+    }
+  }
+}
+
+output "public_certificate_common_name" {
```
junior, line 38: Outputs should be on the outputs.tf

deploy/basic/terraform/loadbalancer.tf
```diff
@@ -71,4 +72,31 @@ resource "oci_load_balancer_listener" "mushop_listener_443" {
...
+    for_each = var.enable_acme_certificate ? [1] : []
+    content {
+      #Required
+      certificate_name = oci_load_balancer_certificate.mushop_certificate[0].certificate_name
+      verify_peer_certificate = false
+    }
+  }
+}
+
+resource "oci_load_balancer_certificate" "mushop_certificate" {
```
junior, line 87: Where is the function to renew the certificate?

deploy/basic/terraform/nsg.tf
```diff
@@ -0,0 +1,1670 @@
...
+  freeform_tags  = local.common_tags
+  vcn_id         = oci_core_virtual_network.mushop_lb_vcn[0].id
+}
+
+resource oci_core_network_security_group_security_rule waf_nsg_rule_1 {
+  destination_type          = ""
+  direction                 = "INGRESS"
+  network_security_group_id = oci_core_network_security_group.waf_nsg.id
+  protocol                  = "6"
+  source                    = "192.157.18.0/24"
```
junior, line 13: Should be a variable or a local. Also does not follow the MuShop CIDR blocks

deploy/basic/terraform/nsg.tf
```diff
@@ -0,0 +1,1670 @@
+resource oci_core_network_security_group waf_nsg {
```
junior, line 1: if this is specific for WAF, the file should be renamed to war-nsg.tf

deploy/basic/terraform/nsg.tf
```diff
@@ -0,0 +1,1670 @@
...
+    }
+  }
+}
+
+resource oci_core_network_security_group_security_rule waf_nsg_rule_81 {
+  destination_type          = ""
+  direction                 = "INGRESS"
+  network_security_group_id = oci_core_network_security_group.waf_nsg.id
+  protocol                  = "6"
+  source                    = "130.35.112.0/22"
```
junior, line 1293: Where these IP addresses are coming from?

deploy/basic/terraform/policies.tf
```diff
@@ -54,6 +54,6 @@ locals {
   ]
   allow_group_manage_local_peering_statements = [
     "Allow group ${var.user_admin_group_for_vault_policy} to manage local-peering-gateways in compartment id ${var.compartment_ocid}",
-    "Allow group ${var.user_admin_group_for_vault_policy} to manage local-peering-gateways in compartment id ${(var.lb_compartment_ocid != "") ? var.lb_compartment_ocid : var.compartment_ocid}"
+  #  "Allow group ${var.user_admin_group_for_vault_policy} to manage local-peering-gateways in compartment id ${(var.lb_compartment_ocid != "") ? var.lb_compartment_ocid : var.compartment_ocid}"
```
junior, line 57: Why commenting? The user need to have a option to have the policy to create LPGs

deploy/basic/terraform/providers.tf
```diff
@@ -25,6 +25,9 @@ terraform {
       source  = "hashicorp/tls"
       version = "2.0.1" # Latest version as March 2021 = 3.1.0. Using 2.0.1 (April, 2020) for ORM compatibility
     }
+    acme = {
+      source = "vancluever/acme"
```
junior, line 29: Need to check is acme is valid for ORM and include the supported version. Also, need to update the .terraform.lck

deploy/basic/terraform/schema.yaml
```diff
@@ -70,6 +70,31 @@ groupings:
     - object_storage_mushop_media_compartment_ocid
     - object_storage_mushop_media_visibility
 
+  - title: "Advanced Resource Options - DNS"
```
junior, line 73: These field should only be visible if advanced options is selected

deploy/basic/terraform/schema.yaml
```diff
@@ -70,6 +70,31 @@ groupings:
...
+    variables:
+    - enable_dns_zone
+    - dns_zone_name
+    - dns_lb_entry
+
+  - title: "Advanced Resource Options - Certificate Generation"
```
junior, line 79: These field should only be visible if advanced options is selected

deploy/basic/terraform/schema.yaml
```diff
@@ -70,6 +70,31 @@ groupings:
...
+    variables:
+    - enable_acme_certificate
+    - acme_email
+    - acme_server_url
+
+  - title: "Advanced Resource Options - Web Application Firewall"
```
junior, line 85: These field should only be visible if advanced options is selected

deploy/basic/terraform/variables.tf
```diff
@@ -192,3 +192,80 @@ locals {
...
+variable "enable_acme_certificate" {
+  default = false
+}
+
+variable "acme_email" {
+  default = ""
+}
+
+variable "acme_server_url" {
+  default = "https://acme-staging-v02.api.letsencrypt.org/directory"
```
junior, line 229: Using staging for production?

images/basic/00-Topology-v1.2.0.MSZ-demo.svg
```diff
@@ -1,3 +1,3 @@
...
...4LjRsMzctMTUuOWwzNywxNS45VjgzLjF6Ii8+JiN4YTsJPHBvbHlnb24gZmlsbD0iIzMxMkQyQSIgcG9pbnRzPSI5OS41LDY5LjYgOTUuMSw3NCA5MC43LDY5LjYgODcuNyw3Mi43IDkyLjEsNzcuMSA4Ny43LDgxLjQgOTAuNyw4NC41IDk1LjEsODAuMSA5OS41LDg0LjUgICAgMTAyLjYsODEuNCA5OC4yLDc3LjEgMTAyLjYsNzIuNyAgIi8+JiN4YTsJPHBvbHlnb24gZmlsbD0iIzMxMkQyQSIgcG9pbnRzPSIxMDQuMSw1Mi45IDEwMS4xLDQ5LjggOTIuNiw1OCA4OS41LDU0LjMgODYuMiw1Ny4xIDkyLjMsNjQuMyAgIi8+JiN4YTsJPHBvbHlnb24gZmlsbD0iIzMxMkQyQSIgcG9pbnRzPSIxMDQuMSw5MiAxMDEuMSw4OC45IDkyLjYsOTcuMSA4OS41LDkzLjQgODYuMiw5Ni4yIDkyLjMsMTAzLjQgICIvPiYjeGE7CTxyZWN0IHg9IjU3LjgiIHk9Ijc0LjUiIGZpbGw9IiMzMTJEMkEiIHdpZHRoPSIyMS4yIiBoZWlnaHQ9IjYuNiIvPiYjeGE7CTxyZWN0IHg9IjU3LjgiIHk9Ijk0LjEiIGZpbGw9IiMzMTJEMkEiIHdpZHRoPSIyMS4yIiBoZWlnaHQ9IjYuNiIvPiYjeGE7CTxyZWN0IHg9IjU3LjgiIHk9IjU0LjkiIGZpbGw9IiMzMTJEMkEiIHdpZHRoPSIyMS4yIiBoZWlnaHQ9IjYuNiIvPiYjeGE7PC9nPiYjeGE7PC9zdmc+" preserveAspectRatio="none"/></g><switch><g requiredFeatures="http://www.w3.org/TR/SVG11/feature#Extensibility"/><a transform="translate(0,-5)" xlink:href="https://www.diagrams.net/doc/faq/svg-export-text-problems" target="_blank"><text text-anchor="middle" font-size="10px" x="50%" y="100%">Viewer does not support full SVG 1.1</text></a></switch></svg>
```
junior, line 3: What changed?

junior (changes requested):

deploy/basic/README_MSZ.md
```diff
@@ -21,10 +21,37 @@ If using local Terraform or on CloudShell:
 - Run `terraform init` to init the terraform providers
 - Run `terraform apply` to create the resources on OCI
 
-You can also include Web Application Firewall (WAF), DNS or other components as desired:
+You can also include Web Application Firewall (WAF), DNS or other components as desired.  Note that a public certificate will be required for WAF, and thus requires a DNS zone be created.  See below for more specific instructions.
```
jeevanjoseph, line 24: WAF will require a domain name, which is a tall order for most people who are playing around with OCI. This should be optional.

deploy/basic/README_MSZ.md
```diff
@@ -21,10 +21,37 @@ If using local Terraform or on CloudShell:
...
+
+Lastly when enabling WAF, you will experience a longer terraform run time than before - it takes roughly 25 minutes to create a WAF Policy, and only after that will you see the WAF DNS record in the created zone.  In order to allow terraform to run and complete without risking a timeout, you should run it as follows:
+```bash
+prompt> nohup terraform apply -auto-approve >> apply.txt 2>&1 &
+prompt> tail -f apply.txt
+```
+
+Once the terraform completes, the WAF Policy and DNS Zone should reflect correctly, and you can hit the external URL, such as `store.mushop.example.com`.  The browser should indicate a clean certificate, and the information dialog for it should show what is configured.  Finally, if the CAPTCHA challenge for WAF is enabled, test that by navigating the store to any product page, and it will challenge you before displaying content.
```
jeevanjoseph, line 53: This can potentially break the load test scripts for the load test and auto-scaling labs.

deploy/basic/terraform/dns.tf
```diff
@@ -0,0 +1,51 @@
...
+    domain = local.dns_lb_domain
+    rdata = lookup(oci_load_balancer_load_balancer.mushop_lb.ip_address_details[0], "ip_address")
+    rtype = "A"
+    ttl = 30
+  }
+}
+
+resource "oci_dns_rrset" "mushop-waf" {
+  # Conditional
+  count = var.enable_dns_zone&&var.enable_waf ? 1 : 0
```
jeevanjoseph, line 34: This should also depend on free-tier `var.use_only_always_free_eligible_resources`. WAF is not included in the Always Free tier presently.

deploy/basic/terraform/dns.tf
```diff
@@ -0,0 +1,51 @@
+resource "oci_dns_zone" "mushop_dns_zone" {
+  # Conditional
+  count = var.enable_dns_zone ? 1 : 0
```
jeevanjoseph, line 3: This should also depend on free-tier `var.use_only_always_free_eligible_resources`. WAF is not included in the Always Free tier presently.

deploy/basic/terraform/dns.tf
```diff
@@ -0,0 +1,51 @@
...
+  #Required
+  compartment_id = var.lb_compartment_ocid
+  name = var.dns_zone_name
+  zone_type = "PRIMARY"
+
+}
+
+resource "oci_dns_rrset" "mushop-internal" {
+  # Conditional
+  count = var.enable_dns_zone ? 1 : 0
```
jeevanjoseph, line 14: I recommend that you resolve this in to a local, based on the `var.enable_dns_zone` and `var.use_only_always_free_eligible_resources`. The stack has a feature to let people use only always free resources when deploying, which this might break.

deploy/basic/terraform/providers.tf
jeevanjoseph: `vancluever/acme` is not a supported provider. https://docs.oracle.com/en-us/iaas/Content/ResourceManager/Concepts/providers.htm

deploy/basic/terraform/network.tf
```diff
@@ -37,7 +37,8 @@ resource "oci_core_subnet" "mushop_lb_subnet" {
   cidr_block                 = lookup(var.network_cidrs, (var.create_secondary_vcn ? "LB-SUBNET-REGIONAL-CIDR" : "MAIN-LB-SUBNET-REGIONAL-CIDR"))
   display_name               = "mushop-lb-${random_string.deploy_id.result}"
   dns_label                  = "mushoplb${random_string.deploy_id.result}"
-  security_list_ids          = [oci_core_security_list.mushop_lb_security_list.id]
+  # If WAF is to be used, do not use the security list - instead the NSG will be applied
+  security_list_ids          = var.enable_waf ? null : [oci_core_security_list.mushop_lb_security_list.id]
```
jeevanjoseph, line 41: Previous comment about conditionals. The Always Free tier needs to be accounted for, as MuShop basic is primarily targeted at Always Free Tier.

deploy/basic/terraform/schema.yaml
jeevanjoseph: This field is mutually exclusive with the "Use only Always Free Resources" option.

deploy/basic/terraform/schema.yaml
jeevanjoseph: This field is mutually exclusive with the "Use only Always Free Resources" option.

jeevanjoseph (changes requested):

jeevanjoseph (changes requested):
This is an excellent set of feature additions. Couple of questions to consider.

- Why make these changes on the Basic version ? The basic version needs to support the Always Free tier and all features that are not in the Always Free needs to have toggles triggered by the high level option to simply `Use Always Free Only`. 
- The primary and recommended way for customers to deploy this is through ORM, so the providers we use needs to be supported ones. 
- There are several "Offline" activities (DNS setup etc.) that don't fit well within the ORM stack deployment model

deploy/basic/terraform/nsg.tf
```diff
@@ -0,0 +1,1670 @@
...
+  source                    = "134.70.92.0/22"
+  source_type               = "CIDR_BLOCK"
+  stateless                 = "false"
+  tcp_options {
+    destination_port_range {
+      max = "443"
+      min = "443"
+    }
+  }
+}
```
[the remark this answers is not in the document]
agregory999: Yes - the product docs cover the (ever-changing) list of WAF endpoints.  In order to force all traffic through WAF, the recommendation is to use the NSG only.  The other repo has more information on the IP list that is used, but I essentially took those rules - I think there is some effort to make this list of IPs more automated and easy to work with, but for now I would need to update it when it is changed.

deploy/basic/terraform/variables.tf
agregory999: I did some tests with CAPTCHA and it appeared to only ask me once per browser session.  In reality I could have chosen any advanced feature to try.

deploy/basic/README_MSZ.md
agregory999: Will fix this and make it more obvious

deploy/basic/terraform/schema.yaml
agregory999: I will update the schema so that all of the DNS and WAF features are not available if free resources is checked off.

deploy/basic/terraform/variables.tf
agregory999: I could switch that so the staging one in the variables file needs to be commented out, but Lets Encrypt uses recommends using non-prod until you are ready, due to API usage limits.  I will have a look at it.

images/basic/00-Topology-v1.2.0.MSZ-demo.svg
agregory999: not sure if this was an accidental check-in - your updated picture has what I did.  Maybe not the cert part.

deploy/basic/README_MSZ.md
agregory999: Completely agree - I will make it more obvious in README.  And detail the requirements.

deploy/basic/README_MSZ.md
agregory999: I would say that if WAF is used and CAPTCHA is not enabled by default, the load test should be ok.  Let me look at the defaults. CAPTCHA should be the last option to enable.

deploy/basic/terraform/dns.tf
agregory999: The complete MuShop is already being updated to include some or all of these features, via a different mechanism, and the basic one that is not Kubernetes might be easier to digest.  This is essentially adding the functionality that @junior added to the diagram within the MSZ readme.  

Regarding ORM, I will re-test, and might have to do something differently if ACME isn't going to work.  A shame, because it does go nicely in with OCI DNS and validation of domain ownership.  If the cert cannot go in, then the listener and WAF are harder to use.

Agreed on the offline activity - I will work on the documentation to show how this is required and what to do.

> This is an excellent set of feature additions. Couple of questions to consider.
> 
> * Why make these changes on the Basic version ? The basic version needs to support the Always Free tier and all features that are not in the Always Free needs to have toggles triggered by the high level option to simply `Use Always Free Only`.
> * The primary and recommended way for customers to deploy this is through ORM, so the providers we use needs to be supported ones.
> * There are several "Offline" activities (DNS setup etc.) that don't fit well within the ORM stack deployment model

deploy/basic/terraform/variables.tf
junior: Yes, you can have an option for staging, but if some will be using a FQDN hostname, standing will not generate valid certificates for any browser.

deploy/basic/terraform/dns.tf
junior: You can request new providers on the ORM, that need to be validated and approved by ORM team and Corp Arch. You cannot have an unsupported provider on the basic stack or will break everything else, even if was supposed to only be used from Terraform and not ORM. Currently Terraform does not support conditional provider.

You have an option to create a fork with that solution, or include the optional instructions on the wiki.

junior:
@agregory99 Are you doing changes to fulfill the requests? or want to do a reduced feature set? Without updates, this PR is considered stale and will automatically be closed.

agregory999:
I spent some time removing the cert gen part, essentially requiring you to have one already, then define it in the variables.  That would be required for merge unless Marketplace supports the acme provider.  Will contact you on Slack when ready and perhaps we can do this together.

junior:
> I spent some time removing the cert gen part, essentially requiring you to have one already, then define it in the variables. That would be required for merge unless Marketplace supports the acme provider. Will contact you on Slack when ready and perhaps we can do this together.

In the past, I was planning to include the certbot on the instance, so would be able to get the lets encrypt cert direct to the compute, after deployed. The only drawback is that the Always-free resource does not include nat gateway. So, if the user select to have private visibility, would not be able to access the internet to install the certbot.

One option is to have a rule that only allow the WAF+CERT+DNS, if no Always-free resources are selected. I can include that feature to facilitate that if you think would help.


Another option is to request the OCI Resource Management team to include the cert provider on the allowed list

PR #274, opened 2021-06-08
Bugfix storefront dev mode

mvandervliet:
- Resolves problems with `make up` in development
- Adds `assets` service to the local development docker-compose


src/storefront/VERSION
```diff
@@ -1 +1 @@
-2.2.2
\ No newline at end of file
+2.2.3
```
junior, line 1: Short lived 2.2.2

junior (approved):

src/storefront/VERSION
mvandervliet: Could have been shorter 😉

PR #282, opened 2021-07-09
Fixes: #281 Feature/pseudo l11n support

### LINKED ISSUE #281: Mushop to accept Australia Adress / States
Issue we are facing currently with Mushop accepts orders only if the address is US
and doesn't have option to accept Australia .. need help to get Mushop to accept users with Australia address

mvandervliet:
Supports country code selection for OCI Regions

```text
option(value="USA") USA
option(value="AUS") AUS
option(value="BRA") BRA
option(value="CHL") CHL
option(value="JPN") JPN
option(value="NLD") NLD
option(value="SAU") SAU
option(value="KOR") KOR
option(value="CHE") CHE
option(value="ARE") ARE
option(value="GBR") GBR
```

Fix #281


junior (approved):
Why OCI regions for the shipping Countries? Just a random convention?

mvandervliet:
> Why OCI regions for the shipping Countries? Just a random convention?

Just to include a handful of likely options for the audience, including the immediate ask for AUS. To do the complete list, I'd prefer moving this to a KV map in the pug data and updating the template to iterate the full list.

PR #286, opened 2021-07-27
Versions update

junior:
Updates the TF providers versions
Updates the Cluster Utilities Helm charts versions
Change the default behavior of the Functions to False and note that only supports the Ashburn-1 region (Temporary workaround until next drop that will support all regions for Functions. Limitation on the Oracle Function deployment)


deploy/complete/terraform/mushop-variables.tf
```diff
@@ -51,8 +51,8 @@ variable "create_oci_service_user" {
   description = "Creates OCI Service User. Service user is needed for Email Delivery (Newsletter feature) and Stream services."
 }
 variable "newsletter_subscription_enabled" {
-  default     = true
-  description = "Enables newsletter subscription feature. Deploys API Gateway, Newsletter Function and uses Email Sender service"
+  default     = false
+  description = "(Currently only supported on the US-Ashburn-1 region) Enables newsletter subscription feature. Deploys API Gateway, Newsletter Function and uses Email Sender service"
```
mvandervliet, line 55: for my curiosity... why only ashburn?

mvandervliet (approved):

deploy/complete/terraform/mushop-variables.tf
junior: To deploy a function from a published image, the image need to be on the same Region. That's Fn limitation.

Issue #291, opened 2021-09-03
Grafana - Local Port Forwarding into internal service


simon-neto:
The installation of the helm chart returned the following output:

enable port forward
 kubectl port-forward -n mushop-utilities svc/mushop-utils-grafana 3000:80

access grafana via the url:
http://localhost:3000/grafana

but I receive the error message "too many redirects" on the browsers I tested (Firefox, Chrome, Safari).

When using the external load balancer URL from the service mushop-utils-ingress-nginx-controller it works.

I think is just a matter of updating the documentation and the helm chart output.

Can you explain if anything has changed on the connection trough the service?

Previously you would have:
kubectl port forward > workstation > mushop-utils-grafana (ExternalName service) > pod

Now we have 
workstation > mushop-utils-ingress-nginx-controller (Load Balancer) > mushop-utils-grafana ( ExternalName service) > pod

junior:
That's an old forgotten update to the notes helm. I'll be updating on the next release.

For now, use the command:
```bash
kubectl port-forward -n mushop-utilities svc/mushop-utils-ingress-nginx-controller 4000:80
```

This will give you access to the Grafana using `http:/localhost:4000/grafana` and the Storefront using `http://localhost:4000/`

junior:
The above comment will only work after you deploy the MuShop chart, which creates the ingress.

I've also updated the Grafana configurations to allow the port-forward direct to Grafana from the service, in case need to access even if just deployed the mushop-utilities....
....so the command `kubectl port-forward -n mushop-utilities svc/mushop-utils-grafana 3000:80` will also work

simon-neto:
Thanks!!! I managed to get it working with the public IP. I just placed the issue to keep record of the documentation in the future.
Thanks Adao

junior:
PR #293 fix the issue

PR #293, opened 2021-09-03
Updates

junior:
Complete:

- Update Grafana notes for port-forward option
- Update Grafana config to allow run from svc directly and not only from the ingress
- Update supported versions for Cluster Autoscaler

Basic:

- Update Terraform OCI provider version
- Update Storage resource with the TF OCI provider update

Fix Issue #291

### DIFF
--- a/src/docs/layouts/shortcodes/content/setup.md
+++ b/src/docs/layouts/shortcodes/content/setup.md
@@ -1,3 +1,5 @@
+#
+
 MuShop provides an umbrella helm chart called `setup`, which includes several
 _recommended_ installations on the cluster. These represent common 3rd party
 services, which integrate with Oracle Cloud Infrastructure or enable certain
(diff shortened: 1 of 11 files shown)


src/docs/layouts/shortcodes/content/setup.md
```diff
@@ -39,13 +41,22 @@ From `deploy/complete/helm-chart` directory:
...
       --namespace mushop-utilities
     ```
->  **OPTIONAL** The Jenkins automation server can be enabled by setting `jenkins.enabled` to `true` in `values.yaml` or by adding the command line flag `--set jenkins.enabled=true` in the `helm install` command above.
 
-  ```shell--helm3
-  helm install mushop-utils setup \
-    --namespace mushop-utilities \
-    --set jenkins.enabled=true
-  ```
+    > **OPTIONAL** If will be provisioning ATP, Stream and Object Storage with OCI Service Broker, you should enable the Service Catalog  `catalog.enabled` to `true` in `values.yaml` or by adding the command line flag `--set catalog.enabled=true` in the `helm install` command above.
```
jeevanjoseph, line 45: If you would like to use the OCI service broker to provision ATP, Stream and Object Storage, you should enable...

jeevanjoseph (approved):
Some optional nitpicks, but LGTM.

src/docs/layouts/shortcodes/content/setup.md
junior: 😱

PR #444, opened 2024-10-02
Adding Disaster Recovery Support to Mushop Catalogue

naikvenu:
This change aims to support Mushop disaster recovery usecase. Only doing this for catalogue service due to effort and complexity involved with other services like carts, orders, user. showcasing this capability should serve the immediate purpose.

- The code changes are done for catalogue: added a failover condition for primary and standby.
- In order to achieve this use case we need to rely on TNS_ADMIN and the sqlnet.ora needs to be changed as below:

WALLET_LOCATION = (SOURCE = (METHOD = file) (METHOD_DATA = (DIRECTORY="?/network/admin")))
SSL_SERVER_DN_MATCH=yes
change to,
WALLET_LOCATION = (SOURCE = (METHOD = file) (METHOD_DATA = (DIRECTORY="${TNS_ADMIN}")))

This is a manual step that will be added to the doc for users who wants to try the DR use case.

- All the values for catalogue are separated and shown as an example under: values-dr.yaml

- Under deploy/complete/helm-chart/mushop/charts/catalogue/values.yaml , i need your suggestion whether we need to separate this image as :dr tag or it needs to be the latest build itself. If it needs to go in the latest build then we need the wallet changes to be done for catalogue. I will provide the steps on that.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -2,33 +2,50 @@ name: Publish Docs
 
 on:
   push:
-    branches: 
+    branches:
     - master
     - docs_update
     paths:
     - 'src/docs/**'
+  workflow_dispatch:
 jobs:
-  build-and-deploy:
+  build:
     runs-on: ubuntu-latest
+    env:
+      HUGO_VERSION: 0.122.0
     steps:
     - name: Checkout
-      uses: actions/checkout@v2
+      uses: actions/checkout@v4
     - name: Setup Node
-      uses: actions/setup-node@v2
+      uses: actions/setup-node@v4
       with:
-        node-version: '14'
+        node-version: ''
     - name: Setup Hugo
-      uses: peaceiris/actions-hugo@v2
-      with:
-        hugo-version: '0.82.0'
-        extended: true
+      run: |
+        wget -O ${{ runner.temp }}/hugo.deb https://github.com/gohugoio/hugo/releases/download/v${HUGO_VERSION}/hugo_extended_${HUGO_VERSION}_linux-amd64.deb \
+        && sudo dpkg -i ${{ runner.temp }}/hugo.deb
+    - name: Setup Pages
+      id: pages
+      uses: actions/configure-pages@v5
     - name: Build Hugo Docs
       run: |
           cd src/docs
           make build
-    - name: Publish to Github Pages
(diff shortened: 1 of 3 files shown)


src/catalogue/cmd/cataloguesvc/main.go
```diff
@@ -109,19 +110,37 @@ func main() {
 	}
 
 	// Data domain.
+	/ Try connecting to the primary database
 	db, err := sqlx.Open("godror", *connectString)
 	if err != nil {
-		logger.Log("err", err)
-		os.Exit(1)
+		logger.Log("Error", "Failed to open Primary Database connection", "CONNECTSTRING", *connectString)
```
junior, line 116: Should not call primary on the logs as can cause confusion.

src/catalogue/cmd/cataloguesvc/main.go
```diff
@@ -109,19 +110,37 @@ func main() {
...
 	}
-	defer db.Close()
 
-	// Check if DB connection can be made, only for logging purposes, should not fail/exit
+	// Test the connection to the primary database
 	err = db.Ping()
 	if err != nil {
-		logger.Log("Error", "Unable to connect to Database", "CONNECTSTRING", connectString)
+		logger.Log("Error", "Unable to connect to Primary Database", "CONNECTSTRING", *connectString)
```
junior, line 122: Same here. Also, the log should show that tried to the main DB and then is trying the standby.

junior (changes requested):
Main issues:
- Oracle client downgraded
- hardcoded wallet version location on multiple places
- Code changes on the service should com first and tested before Kubernetes changes
- Same service is reused by different applications (monolith and micro service)
- DR is not the main use case

naikvenu:
> Main issues:
> 
> * Oracle client downgraded
> * hardcoded wallet version location on multiple places
> * Code changes on the service should com first and tested before Kubernetes changes
> * Same service is reused by different applications (monolith and micro service)
> * DR is not the main use case
> * Service image location should be the common location
> * Service should be versioned, and the build is automated and pushed to the registry. (You can use your fork and configure to push to your container registry before approval and sent to the main)

I have made changes in such a way that both normal and DR will work together. Now, with these changes i am handling the DB initialization part as well. there was a flag missing in catalogue-job.yaml which i have added.

PR #448, opened 2024-10-15
Adding catalogue service changes for DR usecases

naikvenu:
This change aims to support Mushop disaster recovery usecase. Only doing this for catalogue service due to effort and complexity involved with other services like carts, orders, user. showcasing this capability should serve the immediate purpose.

The code changes are done for catalogue: added a failover condition for primary and standby.


naikvenu:
This is very complicated enhancement and breaks at various levels. It's best to close it for now.

